PR #13, opened 2018-04-27
Setup connection

stijnvanhoey:
Update the connection functionality and provide info about which view or table for each of the statements


stijnvanhoey:
merging myself...

PR #14, opened 2018-04-27
Project listing

stijnvanhoey:
Provide the project listing function. The support input check is ok for the `get_projects` function, but not for the other with vector inputs. Although, we can merge now and adapt the functionality further from master.


damianooldoni:
Yes, @stijnvanhoey . I will work on case where `length(arg) > 1` in a new branch.

PR #32, opened 2018-04-30
Quick fix nse

stijnvanhoey:
This provides some quick fixes missed earlier:
- NSE evaluation versus variable overscoping. Fixed by using the `.data` construction and loading this into the namespace
- connnection creation as part of the examples
- fix on con versus connection


stijnvanhoey:
merging directly, as required to have working code ;-)

PR #39, opened 2018-05-02
Frame to tibble

stijnvanhoey:
Converts the outputs to tibbles instead of data.frames


damianooldoni:
@stijnvanhoey: we forgot to merge this pull request after positive review long time ago. I pulled the `master` in this branch, I solved small conflicts and run `devtools::build_vignettes()` and no problems occurred. Could you please check it? This time we mustn't forget to merge :smile:

stijnvanhoey:
did you perform a package check `devtools::check(), or press Ctrl/Cmd + Shift + E in RStudio`?

damianooldoni:
No. Indeed, there are very small things to fix still. Thanks.

damianooldoni:
`devtools::check()` returned 5 errors, which are actually test failures. 
I could easily solve three of them (see last three commits: af14072,  79c4eba, 335a685).
However, I have no idea about the two failed test shown here below, both concerning column `recover_date_time`:
```r
test_get_deployments.R:75: failure: test_output_get_deployments
all(test2 %>% select(recover_date_time) %>% is.na()) isn't true.

test_get_deployments.R:77: failure: test_output_get_deployments
nrow(test6 %>% filter(is.na(recover_date_time))) not equal to nrow(test2).
1/1 mismatches
[1] 1 - 28 == -27
```
@stijnvanhoey: Do we actually need these two tests?

stijnvanhoey:
yes, we do need the tests (or rewrite them), as the `recover_date_time` column is [used](https://github.com/inbo/etn/blob/master/R/get_deployments.R#L69) in the function `get_deployments` to distinguish active and non-active deployments. 

Maybe not connected with the frame/tibble changes, but requires check. 

Certainly this check makes sense:
```
test2 <- get_deployments(con, network_project = "thornton")
test6 <- get_deployments(con, network_project = "thornton",
                         open_only = FALSE)
expect_equal(nrow(test6 %>% filter(is.na(recover_date_time))), nrow(test2))
```
`open_only`, i.e. open neworks means that there is no  `recover_date_time` yet defined (not taken back), so adding the parameter or filter on it later on should provide the same result! 

@bwydoogh anything changes on the dbase side according to the `recover_date_time` characteristics?

damianooldoni:
All bugs solved. Especially the last one about `open_only` (5b34946) :+1: 
@stijnvanhoey: could you please double check before merging? Thanks.

stijnvanhoey:
thanks Damiano, nice work!

Issue #38, opened 2018-05-01
Add information to potential etn users about setup/login info/...


stijnvanhoey:
As the package only works within the Rstudio server (or within the VLIZ network), we have to provide this information to potential users in the [readme](https://github.com/inbo/etn/blob/master/README.md) (which is the landing page of the documentation). 

@jreubens could you add that kind of information (here or direclty edit the [readme](https://github.com/inbo/etn/blob/master/README.md))? Short the concept ( and link to other pages), how they can get a login, what the data can be used for, some other useful links,...

jreubens:
when connecting to ETN I had to add a comment line to read my .Renviron file:
'readRenviron("~/.Renviron")'
Without this comment line I could not access my credentials

Thanks to Benny (and google) to sort this out

stijnvanhoey:
Thanks for solving this out. We will add this to the documentation.

jreubens:
I would write the following:

'This package is built for users from the European Tracking Network (ETN) http://www.lifewatch.be/etn/
It provides access to data and metadata stored in the ETN central database.
This package only works within the Rstudio server of VLIZ (http://rstudio.lifewatch.be/) and is password protected. A login can be requested at http://www.lifewatch.be/etn/contact.
You will be provided with credentials which are needed to connect to the database.
Once connected you can access the ETN (meta)data. Access to data is linked to your user profile and to the moratorium rules implemented in the ETN data policy (http://www.lifewatch.be/etn/assets/docs/ETN-DataPolicy.pdf?1.0). '

@PieterjanVerhelst, @IPauwels agree with this explanation as frequent users? Feel free to adapt...

peterdesmet:
I would provide a step-by-step instruction how to request access. Adapting from Jan's text:

> This package provides functionality to access and process data from the [European Tracking Network (ETN)](http://www.lifewatch.be/etn/) database hosted by the Flanders Marine Institute (VLIZ) as part of the Flemish contribution to LifeWatch. ETN data is subject to the [ETN data policy](http://www.lifewatch.be/etn/assets/docs/ETN-DataPolicy.pdf) and can be restricted (under moratorium and only accessible to data owners/collaborators) or unrestricted (public and routinely published to international biodiversity facilities).
> 
> The ETN infrastructure currently requires the package to be run within the [LifeWatch.be RStudio server](http://rstudio.lifewatch.be/), which is password protected. To request access:
> 
> 1. Request a login at http://www.lifewatch.be/etn/contact
[...]

PR #51, opened 2018-05-28
Update readme

stijnvanhoey:
Updates the text with the agreed intro defined in #38


stijnvanhoey:
Merging myself, only documentation updates

Issue #45, opened 2018-05-08
get_deployments


jreubens:
When you filter for 'Active' deployments for a specific project, you currently get a list of all deployments that have occured  in this project for the (currently) active receivers. This includes also closed deployments.
What we want is a list of the 'open deployements (i.e. no end date). thus probably we need to include an extra criterion in this function.
@PieterjanVerhelst  correct me if I'm wrong

jreubens:
Currently you can only filter for a specific network_project.
Would be nice if you could also filter for a specific station_name

jreubens:
indead :-)
I changed the name of the issue...

PieterjanVerhelst:
@jreubens wouldn't it be interesting to make the selection on open deployments optionally? It can be handy to see the whole track of deployments for a specific station.

jreubens:
I would suggest:

- if you filter for a project you get the open deployments by default

- if you filter for a station you get all deployments by default

But in both options you can change settings, depending on your needs
agree?

PieterjanVerhelst:
Based on #43 , I suggest to just keep the function as it is, returning all deployments. Extracting the open deployments can be done by, for instance, a separate pipe.

jreubens:
Based on #43 I agree that we should keep the functionality as it is.
Regarding the filtering for 'active' deployments. I'm still convinced you should only get an overview of the 'open' deployments (i.e. no end date). In this filtering you are not interested in all deployments that occurred for stations that have an 'active' deployment.
For the latter you can use an extra pipe in dplyr...

Adding the open deployments in the filter allows you to quickly have an overview af where you have receivers active at the moment. This will often be used as input to decide which receivers need maintenance or have a battery replaced.

stijnvanhoey:
@jreubens can you explicitly define the column name and the value to filter on (or if it needs more requests, define in detail what condition an active deployment define).

Am I correct that it is those records with an `NA` value for `recover_date_time` (i.e. no end date):
```
get_deployments(my_con) %>% filter(is.na(recover_date_time))
```
can you have a check?

jreubens:
This is indeed correct!

PR #53, opened 2018-05-29
Active versus open  deployments

stijnvanhoey:
This PR handles the request about open deployments defined in #45. By default, the open deployments (no end date, `open_only = TRUE`) are shown, but the user can define `open_only` to FALSE to get all deployments.


jreubens:
@stijnvanhoey  _'but the user can define open_only to get all deployments'_
Isn't is more logical to call this 'all_deployments' instead of open_only... 
with the latter I would expect to have only the open deployments. Or do I see something in the wrong way?

stijnvanhoey:
@jreubens as it is implemented now, the default action is to have the open deployments (no end date) by default (I called this `open_only = TRUE`), but the user can override this by putting `open_only` on FALSE...

jreubens:
ok, now I understand:-) my mistake
great job!

Issue #43, opened 2018-05-08
my_animals


PieterjanVerhelst:
Some bugs in the `get_animals()` functionality:
- When I run the code `my_animals <- get_animals(my_con, animal_project = "2012_leopoldkanaal")` I get the animals for both the `2012_leopoldkanaal` and `PhD van der Knaap` projects. This also holds true when other animal projects are selected; apparently `PhD van der Knaap` comes as a standard in the output.

- The function does not return information about `length_type3` and `length_type4` nor the actual weight of the fish.

stijnvanhoey:
With regard to adding additional filtering options in this issue as well as some other issues  (e.g. #45, #46), make sure it is worthwhile to add the functionality. For `get_detections`, these filters provide the ability to limit the requested data and overcome the risk of getting too large tables and large useless  data transfers. 

However, for functions that will return just a relative small amount of data (even without the filter), we maybe should not overload the function arguments, but just provide documentation on how to filter the data after requesting the data, e.g. just one `dplyr` function call away to subset the owner organisation:
```r
my_animals <- get_animals(my_con, animal_project = "2012_leopoldkanaal") 
my_animals_inbo <- my_animals %>% filter(owner_organisation == "INBO")
```
the key point is the table size requested (for `detections` typically large, for animals much smaller).

As such, in order to not end up with tons of arguments for each function that are (1) harder to maintain and (2) create overload for the functions; let's make sure to make this a bit more efficient as in 'would be nice':
* for each of the additional filter request, think about the potential data size decrease to load from the database. If this is relatively small or if the total data size will never be huge; the added value is minimal compared to just adding a filter statement with `dplyr` after requesting the data. 
* If it is useful to add the filter, make sure to provide the following information:
   * make sure you can clarify the added value with a concrete example for the users (`would be nice` is not enough ;-))
    * is the column/information to filter on already in the returned tabel? If yes, define the column. If no, provide the required information that should be added to the view on dbase level or that is available in other tables
    * what is the user expected to provide: single name, list of names, TRUE/FALSE,... ?

@jreubens and @PieterjanVerhelst for those open issues with a would be nice, please do a revision of these requests and make sure the requirements above are valid for each of them. Thanks!

PieterjanVerhelst:
I agree; comments added to the issues. The only function where additional filters can still be helpfull, is the `get_detections()` function. So the more data that can be reduced on the data extraction level, the better. See examples at #47 .

JolienGoossens:
[...]

Issue #41, opened 2018-05-07
get_receivers


jreubens:
currently the get_receivers function links receivers to a network.
Some remarks on this:
- output gives an overview of all deployment that occur/have occurred. Thus receivers can be present multiple times. I don't think this is a wanted output. Should be unique receives to my opinion
- currently you can only link to the network_project. should be extended to animal_project as well
- next to projects you should also be able to view and filter on owner_organisation. The output gives a number. should be a name.... TO be able to filter on it using 'get_receivers(my_con, network_project = "INBO")' I guess this info needs to be added to 'my_projects' output first...

stijnvanhoey:
@jreubens are these nice to have features are rather required package functionalities to tackle for the first release?

jreubens:
The ones in this issue are more than 'nice to have' thus should be part of the first release to my opinion

damianooldoni:
@stijnvanhoey , @jreubens : nice. I will find the needed time to implement this features and update documentation and tests before the start of my 10 days holidays on Wed 6th of June. :+1:

stijnvanhoey:
I agree on point 

1. receivers should not provide information on all deployments

3. as the `owner_group_fk` can [now](https://github.com/inbo/etn/issues/42#issuecomment-392708047) be linked to the `etn_group`, we can add this to the table

I do not agree on point

2. animal_projects are linked to animals; in order to be able to filter network projects as well, the issue is similar to #43, but now the other way around. For all detections, which receivers got signal from which tags, tags linked to tag_releas, to animal to animal_project... That's a long road to just add this filter... 

It depends on the necessity, but I see the discussion of point 2 similar to #43, so let's discuss this kind of filter there. This issus focuses on the solving of point 1 and 3, a todo for me and damiano.

peterdesmet:
Trying to close this issue.

> 1. Should be unique receives to my opinion

Agree, this is now the case. `network_project_code` is no longer a parameter

> 2. ... currently you can only link to the network_project. should be extended to animal_project as well

As mentioned by @stijnvanhoey, let's not do this: it's a complicated way to get there and doesn't really make sense in the model.

> 3. ...next to projects you should also be able to view and filter on owner_organisation

[...]

PR #54, opened 2018-05-30
Receivers patch

stijnvanhoey:
This PR tackles the discussion of issue #41:
- it removes the duplicate rows of receivers when multiple deployments where done on a single receiver; as this still results in multiple network_projects linked to a single receiver (networks are linked to deployments, hence, when these deployments had different projects, the receiver is linked to multiple network projects), the column network_projectcode provides a combined list of projects, separated by `,`
- is adds the owner_organisation column by joining the etn_group table

as an example (just a subset of the columns and rows):

|serial_number |status    |network_projectcode                     |owner_organisation |
|:-------------|:---------|:-------------------------------|:------------------|
|113528        |Active    |leopold,zeeschelde,bovenschelde |INBO               |
|3110          |Active    |cpod-lifewatch                  |VLIZ               |
|113524        |Active    |dijle,bovenschelde,leopold      |INBO               |
|124072        |Lost      |albert                          |INBO               |
|546193        |Available |MorayFirth                      |AST                |
|129709        |Active    |bpns,Noordzeekanaal             |VLIZ               |

remark, I do NOT
- add the animal_project as additional filter, see #43 discussion
- add the owner_organisation as filter, as this is only one command away, e.g. `get_receivers(con) %>% filter(owner_organisation == "INBO")`

### DIFF
--- a/tests/testthat/test_get_receivers.R
+++ b/tests/testthat/test_get_receivers.R
@@ -28,5 +28,6 @@ testthat::test_that("test_output_get_receivers", {
   expect_equal(ncol(test1), ncol(test2))
   expect_equal(ncol(test2), ncol(test3))
   expect_true("owner_organisation" %in% colnames(test1))
+  expect_true("network_projectcode" %in% colnames(test1))
 })
 
--- a/tests/testthat/test_get_receivers.R
+++ b/tests/testthat/test_get_receivers.R
@@ -22,10 +22,11 @@ testthat::test_that("test_output_get_receivers", {
   expect_is(test1, "data.frame")
   expect_is(test2, "data.frame")
   expect_is(test3, "data.frame")
-  expect_gte(nrow(test1), nrow(test2))
+  expect_gt(nrow(test1), nrow(test2))
   expect_gte(nrow(test1), nrow(test3))
   expect_gte(nrow(test3), nrow(test2))
   expect_equal(ncol(test1), ncol(test2))
   expect_equal(ncol(test2), ncol(test3))
+  expect_true("owner_organisation" %in% colnames(test1))
 })
 
--- a/R/get_receivers.R
+++ b/R/get_receivers.R
@@ -60,6 +60,7 @@ get_receivers <- function(connection,
   receivers %>%
     group_by(id_pk) %>%
     mutate(projectcode = paste(projectcode, collapse = ",")) %>%
+    rename(network_projectcode = projectcode) %>%
     ungroup() %>%
     distinct()
 }
--- a/R/get_receivers.R
+++ b/R/get_receivers.R
@@ -50,5 +50,13 @@ get_receivers <- function(connection,
                               .con = connection
   )
   receivers <- dbGetQuery(connection, receivers_query)
-  receivers
+
+  # we still have multiple records of receivers, as project codes are coupled to
+  # deployments and a receiver can have multiple deployments aka projects.
+  # combine the individual network projects in a single row:
+  receivers %>%
+    group_by(id_pk) %>%
+    mutate(projectcode  = paste(projectcode, collapse =",")) %>%
+    ungroup() %>%
+    distinct()
 }
--- a/R/get_receivers.R
+++ b/R/get_receivers.R
@@ -42,7 +42,7 @@ get_receivers <- function(connection,
   }
 
   receivers_query <- glue_sql("
-      SELECT receivers.* , deployments.projectcode
+      SELECT DISTINCT receivers.* , deployments.projectcode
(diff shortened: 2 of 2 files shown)


tests/testthat/test_get_receivers.R
```diff
@@ -22,10 +22,12 @@ testthat::test_that("test_output_get_receivers", {
   expect_is(test1, "data.frame")
   expect_is(test2, "data.frame")
   expect_is(test3, "data.frame")
-  expect_gte(nrow(test1), nrow(test2))
+  expect_gt(nrow(test1), nrow(test2))
   expect_gte(nrow(test1), nrow(test3))
```
damianooldoni: Can we then coherently change this `expect_gte` to `expect_gt` as well? Not such a big difference at the end of course.

PieterjanVerhelst:
Just to make sure we don't make any mistakes: what the `get_receivers()` functionality should be able to do, is giving an output of unique receivers. As @stijnvanhoey points out, there are several `network_projectcodes` per receiver... So this would mean that we still get the same receivers in the output, but one for each network projectcode. I think this is a correct and necessary implementation. @jreubens how do you see this?

tests/testthat/test_get_receivers.R
```diff
@@ -22,10 +22,12 @@ testthat::test_that("test_output_get_receivers", {
   expect_is(test1, "data.frame")
   expect_is(test2, "data.frame")
   expect_is(test3, "data.frame")
-  expect_gte(nrow(test1), nrow(test2))
+  expect_gt(nrow(test1), nrow(test2))
```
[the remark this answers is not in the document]
stijnvanhoey: for this particular case, it make sense to have `gt`

tests/testthat/test_get_receivers.R
stijnvanhoey: no, should be checked for each, I would leave it as such for now

stijnvanhoey:
Maybe to clarify. Do we want to present the result to the user as follows:
(I removed the other columns for clarity):

|serial_number |status    |network_projectcode |
|:-------------|:---------|:-------------------|
|113528        |Active    |leopold             |
|113528        |Active    |zeeschelde          |
|113528        |Active    |bovenschelde        |
|3110          |Active    |cpod                |
|3110          |Active    |lifewatch           |
|113524        |Active    |dijle               |
|113524        |Active    |bovenschelde        |
|113524        |Active    |leopold             |
|124072        |Lost      |albert              |
|546193        |Available |MorayFirth          |
|129709        |Active    |bpns                |
|129709        |Active    |Noordzeekanaal      |

-> when receivers are linked to multiple network projects, the receiver is repeated and the column network_projectcode provides the different projects

or as such

|serial_number |status    |network_projectcode             |
|:-------------|:---------|:-------------------------------|
|113528        |Active    |leopold,zeeschelde,bovenschelde |
|3110          |Active    |cpod-lifewatch                  |
|113524        |Active    |dijle,bovenschelde,leopold      |
|124072        |Lost      |albert                          |
|546193        |Available |MorayFirth                      |
|129709        |Active    |bpns,Noordzeekanaal             |

-> in order to keep 1 receiver per row, we combine all network_projects in a single field

stijnvanhoey:
@damianooldoni I - personally - to think that in this case the aggregation to a single field is valid :-)

stijnvanhoey:
@jreubens what do you think?

jreubens:
first of all, sorry for my delayed reply. Field season is starting...
Why do we need the network_projectcode here? 
To my opinion in this call you want to know:
- which receivers do I own (serial number)
- what is their status (active, broken, lost,..)

Thus I don't think we need the network_project info here. But we do need it on deployment level!
@PieterjanVerhelst  @IPauwels  agree?

stijnvanhoey:
Removing the `network_projectcode` is possible as well, but notice that you requested the ability to filter on `network_project` for the `get_receiver` function as an input argument. As such, we need the network_code to be linked with the data anyway inside the function and it feels convenient to have it in the output as well (otherwise, it is like blind confidence on the returned data).
Still, it is possible to remove the column and this would *solve* the issue of multiple rows of a single receiver...

P.S  
```
> "projectcode" %in% colnames(get_deployments(con))   
> [1] TRUE 
```
you have it on deployment fucntion output... ;-)

jreubens:
OK, than we keep it in, but in the format you suggested to have all network_projects in one line...

damianooldoni:
Thanks for double checking the need of this column! Column `network_projectcode` is pure informative so no problem from my side anymore :+1:

PieterjanVerhelst:
👍

Issue #47, opened 2018-05-08
get_detections


PieterjanVerhelst:
I think the `get_detections()` functionality can be improved by adding filters on:
- Receiver
- species

jreubens:
I agree... btw transmitter is already included (see ?get_detections)

PieterjanVerhelst:
Indeed, I removed transmitter from the improvement list.

PieterjanVerhelst:
Since retrieving detections is, to my opinion, the core of the package and results in large datasets, adding the above suggested filters may improve the function `get_detections()`. For instance, in the project `2015_phd_verhelst` two species (i.e. `Gadus morhua` and `Anguilla anguilla`). Since we are often interested in one species for analysis, selecting that one species during the 'data call' can reduce the requested data substantially (in this case, for 2017 data: 577570 records for `Gadus morhua` and 286962 records for `Anguilla anguilla`). 
Likewise, it can occurr that one wants to check the efficiency/number of detections for a single receiver. Than it would be more efficient to reduce the extraction on the server level, instead of downloading all data and locally selecting on only a fraction of that data.

PieterjanVerhelst:
I was trying to download the dataset of the animal network `2013_albertkanaal` and apparently we cannot download files larger than 100 Mb. Is it possible to allow download of larger files?
Here is the error I got: `Error: cannot allocate vector of size 100.0 Mb`

JolienGoossens:
In `get_detections` I would change the `deployment_station_name` argument to `station_name` (is called station_name in the dataframes).

stijnvanhoey:
@JolienGoossens Actually, both are in...

```
> colnames(get_detections(con, limit = 10))
 [1] "receiver"                  "transmitter"               "transmitter_name"          "transmitter_serial"       
 [5] "sensor_value"              "sensor_unit"               "sensor2_value"             "sensor2_unit"             
 [9] "station_name"               # STATION_NAME
"datetime"                  "id_pk"                     "qc_flag"                  
[13] "file"                      "latitude"                  "longitude"                 "deployment_fk"            
[17] "scientific_name"           "location_name"             
"deployment_station_name"   # DEPLOYMENT_STATION_NAME
"deploy_date_time"         
[21] "animal_project_name"       "animal_project_code"       "animal_moratorium"         "network_project_name"     
[25] "network_project_code"      "network_moratorium"        "signal_to_noise_ratio"     "detection_file_id"        
[...]

PR #55, opened 2018-05-30
Detections features

stijnvanhoey:
This PR tackles the additional filters as described in #47:
- an additional filter argument for scientific_name (cfr. existing `get_animals` argument)
- an additional filter argument for receivers


PieterjanVerhelst:
When I run the function `get_detections()` with `scientific_name` or `receiver` term, I get an error stating that the argument is not used (I checked the spelling of the animal and receiver name and is correct):
`Error in get_detections(my_con, animal_project = "2015_phd_verhelst",  : 
  unused argument (receiver = "VR2W-112297")`

Notably, the function still works when using the 'old' filters like `animal_project` or `deployment_station_name`.

R/get_detections.R
```diff
@@ -103,6 +114,22 @@ get_detections <- function(connection, network_project = NULL,
...
+    receiver = valid_receivers
+  }
+
+  # valid scientific names
+  valid_animals <- scientific_names(connection)
+  check_null_or_value(scientific_name, valid_animals, "scientific_name")
+  if (is.null(scientific_name)) {
+    scientific_name = valid_animals
+  }
+
```
damianooldoni: nice!

damianooldoni:
Indeed.. I just saw your comment after deleting the branch, then reopened it and redoing your test: `get_detections(my_con, animal_project = "2015_phd_verhelst", receiver = "VR2W-112297")` and indeed everything is ok. But is gooed! Double, triple check is more than welcome! Nice.

Issue #56, opened 2018-05-31
Consistency project_code and project_name


damianooldoni:
While writing a vignette about the usage of project code (instead of project name) (see #44) I found a kind of inconsistency about the name of the columns related to project codes and project names.

Here below the thing I am speaking about.

In functions like `get_detections()` we get columns named like this:

```r
"animal_project_name", "animal_project_code", "network_project_name", "network_project_code"
```
which is fine.

While running functions `get_projects()` we get:
```r
get_projects(my_con) %>% colnames()
 [1] "id"              "name"            "projectcode"     "type"           
 [5] "startdate"       "enddate"         "moratorium"      "imis_dataset_id"
 [9] "latitude"        "longitude"
```
As you see, columns `projectcode` and `name` are returned. I would expect `project_code` and `project_name`. The same holds true for `get_receivers()` which returns a data.frame with a column named `projectcode` while all other columns containing the word `code` contain an underscore: `*_code` like `ar_release_code` and `ar_disable_code`.

damianooldoni:
Another example is the following:
```r
get_animals(con, animal_project = "2012_leopoldkanaal") %>%
  colnames()
[1] "projectname"                              
 [2] "projectmember"                            
 [3] "tag_code_space"                           
 [4] "person_id"                                
 [5] "animal_id"                                
 [6] "scientific_name"                          
 [7] "common_name"                              
 ...
67] "animal_tag_release_id_pk"                 
[68] "thelma_converted_code"                    
[69] "projectcode" 
```

All column names contain words separated by underscore except project-related ones (columns [1], [2] and [69]). Maybe there is a reason for it, but I don't find it.

peterdesmet:
The parameter names now consistently use `under_scores`. Project names have been dropped (only `codes` are used). The project table still returns fields that are not underscore, but that should be fixed at the database site.

Issue #44, opened 2018-05-08
Project names and project codes


PieterjanVerhelst:
It is sometimes not very clear if you need to use a project name or project code in the function. I think for us it is often quite obvious, but for external/new users in the future, this may require some extra documentation.

stijnvanhoey:
We will have to document that we always rely on the `projectcode`, should be added to the documentation.

peterdesmet:
Parameter names have been updated to the more explicit `animal_project_code` and `network_project_code`.

peterdesmet:
#57 closed as well

PR #57, opened 2018-05-31
Add vignette about projectcode

damianooldoni:
I wrote a short tutorial about the possible confusion about project codes and project names as mentioned in #44 .

I tried to tackle the problem in the most easy way. Notice that there is still a subtle inconsistency because sometimes we have `projectname` sometimes just `name` (see #56). But I think the user would understand the concept as well.

The only thing I couldn't do is including the vignette in the html. I tried via `pkgdown::build_site()` but I don't get the page online. I think it is a problem about `pkgdown_sha:`. In fact after a `build_site()` run I get it erased to `pkgdown_sha: ~`. I tried to solve the problem, but I didn't succeded. Maybe you can give me a hint next time we meet? :+1: And maybe can you easily solve it in this PR? Thanks.

Another point: I used `knitr::kable()` as you did in `access-etn-data.Rmd`. I am just wondering if such code line would not be confusing for the typical R user. Actually it is just a functional commando for making the tutorial appearance nicer.

Here below a first idea to make both code chunk and html nice:
- Use `eval = FALSE` in the code chunk **without** `knitr::kable()` at the end
- Use `echo = FALSE` in the same code chunk **with** `knitr::kable()` at the end

If you think I am right but you think it is better to tackle afterwards in a separate branch afterwards, I agree totally with you.


peterdesmet:
Best to include your df print option (e.g. `kable`) in the yaml settings of the Rmd, rather than in the code.

peterdesmet:
Did a quick review, and I have updated the vignette file names and updated their title:

- etn-overview: Introduction to main functions of the ETN package
- access-project-info: Get project information

@damianooldoni can you rebuild the site on the Rstudio server?

@damianooldoni note: the reason why your vignette was not included in the menu, is because you edited `docs/pkgdown.yml`. You have to edit `_pkgdown.yml` in the root, which is the source file. Good rule of thumb: never manually edit anything in docs.

damianooldoni:
Thanks @peterdesmet for your help and suggestions! I rebuild the site.

damianooldoni:
Yes, it works. I rebuilt the site correctly. 
The only thing is the absence of a key now at `pkgdown_sha: ~`.
@stijnvanhoey: does it have something to do with Travis CI?

stijnvanhoey:
@damianooldoni you shouldn't be worried about the sha...  Notice that there is no travis setup for this repo, as we can not provide the data access outside the VLIZ Rstudio server environment.

stijnvanhoey:
and +1 for adding the kable as yaml setting instead of individual chunk settings, cfr.

try with an adapt of the YAML headers:
```
output: rmarkdown::html_vignette
  html_document:
    df_print: paged
```
(we use paged instead of kable)

peterdesmet:
I didn’t check the output, but maybe the dataframes are too small to be paged? Do they look different than regular boring df print?

damianooldoni:
Well, now they are quite similar to the output you get on a console...

stijnvanhoey:
- I've tested adding the config on level of `_pkgdown` (as done in Rmarkdwon webpage)  ->no effect
- adapting  the yaml header options -> no effect
...

peterdesmet:
This vignette is no longer necessary, see https://github.com/inbo/etn/issues/44#issuecomment-589728136

Regarding nicer table printing, we'll have to figure that out for the main vignette anyway, so I'm on it.

Issue #24, opened 2018-04-27
Exclude "sync tag" when returning detections


peterdesmet:
See list in #21. I would propose to exclude "sync tag" by default when returning occurrences, but maybe allow a flag to include those. @jreubens @PieterjanVerhelst are there other _non animal_ detections in the database?

PieterjanVerhelst:
Yes, the `Built-in` and `Sentinel` are non-animals as well.

peterdesmet:
Ok, so for any relevant function, it would be useful something like a `exclude_non_animals` parameter, with default value `TRUE`. Better name suggestions welcome.

jreubens:
Ok to have an function to exclude non-animals.
as name I would go for 'reference_tags' 
@PieterjanVerhelst OK?

peterdesmet:
I'd opt for parameter, not separate functions. Maybe called `include_reference_tags`, default `FALSE`

PieterjanVerhelst:
Reference tags are a good, general name for non-animal tags. Function name proposed by Peter is ok for me.

stijnvanhoey:
For which functions is this relevant?

* [ ] get_projects() | Get project overview
* [ ] get_receivers() | Get receiver data
* [x] get_deployments() | Get deployments data
* [ ] get_animals() | Get animals data
* [x] get_transmitters() | Get transmitter metadata

@PieterjanVerhelst  and  @jreubens can you verify this?

PieterjanVerhelst:
I think it is only relevant for the` get_transmitters()` function. Deployments reflect receiver activity and therefore not relevant for that function.

jreubens:
get_animals as well.
BUT... we will link 'scientific_name' to aphia of worms soon. Thus this means that built-in, sentinel, ... can't be seen as 'animals' any longer. an new field will be needed. @bwydoogh en @jreubens wil work on this

stijnvanhoey:
Regarding issue #50 and the discussion above. Let's summarize the required implementation:

The aim is an additional argument `include_reference_tags` (default `FALSE`), so for example the implementation would be:
```
get_transmitters(con) #no reference tags included
get_transmitters(con, include_reference_tags = FALSE) #no reference tags included
get_transmitters(con, include_reference_tags = TRUE) #reference tags included
```

Still, I do not completely get how to exclude these? On what field name doe we have to filter or how can I extend the query? Moreover, we need a complete and proper list of values to identify *non-animals*.

[...]

PR #61, opened 2018-06-18
Include reference tags parameter to function get_transmitters()

damianooldoni:
Based on discussion in #24, I changed function `get_transmitters()` by adding parameter `include_reference_tags`. 
The function `get_transmitters()` returns only *animal* tags by default unless `include_reference_tags = TRUE`.
Documentation and tests have been updated as well. 
Review is welcome!


damianooldoni:
Done. Still, the commando `pkgdown::build_site()` gives me the following error while trying to read 'vignettes/access-etn-data.Rmd':

```r
Error in group_by(., id_pk) : could not find function "group_by"
Calls: withCallingHandlers ... .f -> render_rmarkdown -> <Anonymous> -> get_result
Execution halted
Error in group_by(., id_pk) : could not find function "group_by"
```

Do you get the same error? I try installing `pkgdown` from GitHub and getting the most updated version, maybe this can solve the problem.

damianooldoni:
Nope. `build_site()` returns the same error again. Any idea? Am I doing something wrong? As I know, the commando `pkgdown::build_site()` is what we always do. That `group_by` has nothing to do with `dplyr::group_by()`, I think.

stijnvanhoey:
updated, merging now

PR #62, opened 2018-08-08
Fix changed transmitter tag code to compare all tags, see #47

stijnvanhoey:
Fixes the changes in database model of etn, as discussed in #47


stijnvanhoey:
As this is critical and just a dbase column change, mergin myself (all unit test were ok)

PR #67, opened 2019-01-07
Adding unit-tests to check column names in db views

damianooldoni:
This PR improves unit-tests. In particular, tests of column names of db views as returned by all `get_*()` functions have been added. 

The unit-tests will throw an error if:
1. A new column has been added
2. A column name has been changed
3. A column has been removed

The unit-tests will NOT throw an error if:
1. the order of the columns has been changed 

Some minor changes about code uniformity and general improvements of unit-tests have been applied as well.


stijnvanhoey:
Great idea! Thanks for the development.

PR #71, opened 2019-07-23
Add Travis

peterdesmet:


peterdesmet:
I will close this PR, as we can't run the package from anything else than rstudio.lifewatch.be. Should be easy to add Travis config once we can. I will fix the `mailto:` in master though

Issue #88, opened 2020-02-19
How to return animal-tag relationship?


peterdesmet:
Currently:

- `get_tags()` doesn't return unique tags, as it's internally linked to animals to allow filtering on animal project. Since a tag (`A69-1303-20695`) can be associated with multiple animals (`673` and `674`), duplicates are returned
- `get_animals()` doesn't return unique animals, as the view return animals + their tag ID. Since an animal (`2369`) can be associated with multiple tags (`A69-9006-971` and `A69-9006-972`) duplicates are returned.

I would:

- `get_tags()`: remove the option to select on animal project, so the function (just like the view and the table on the website) returns **unique tags**
- `get_animals()`: return **unique animals** by default, which is similar to how the website displays a single row per animal:

<img width="856" alt="Screenshot 2020-02-19 at 16 25 34" src="https://user-images.githubusercontent.com/600993/74852124-78ea8500-533c-11ea-8bda-94651cb1a510.png">

- `get_animals()`: add an option to include `tag_fk` (resulting in multiple), so one can join with information from `get_tags()`

damianooldoni:
After discussing `get_animals()` with @peterdesmet, we opt for the following option:  return **unique animals** collapsing tags in a comma separated string.

damianooldoni:
There are multiple tag related columns in returned data.frame of `get_animals()`:
1. `tag_id`
2. `tag_fk`
3. `tagger`
4. `tagging_type`
5. `tagging_methodology`

Example:
| pk | tag_id | tag_fk | tagger | tagging_type | tagging_methodology
| --- | --- | --- | --- | --- | --- |
2827 | A69-9006-904 | 1794 | - | Internal | 
2827 | A69-9006-903 | 1793 | - | Internal | 

A possible strategy is to collapse not only `tag_id` but all `tag*` columns to comma separated values. Example:

| pk | tag_id | tag_fk | tagger | tagging_type | tagging_methodology
| --- | --- | --- | --- | --- | --- |
2827 | A69-9006-904,A69-9006-903 | 1794, 1793 | -,- | Internal,Internal | 

@peterdesmet, what do you think?

peterdesmet:
Agree: all tag* columns

damianooldoni:
PR #106 should solve uniqueness of `get_animals`(). 
About `get_tags()`, it seems that it is already fixed.

peterdesmet:
Fixed in #106

PR #106, opened 2020-03-12
Solve animal tag relationship

damianooldoni:
This PR solves #88.
Function `get_animals()` updated, its documentation and its unit-tests.

Other small changes: function `as_tibble` imported from `dplyr` instead of `tibble` to reduce `etn` package's dependencies.


damianooldoni:
I will now add a test about the collapsing of multiple values in `tag*` columns. Now there is only a test about uniqueness of animals (`pk`).

damianooldoni:
Done.

peterdesmet:
@damianooldoni can you explain why we need purrr, stringr and tidyselect?

damianooldoni:
In order to test over all columns where collapsing can occur, i.e. all `tag*` columns.
See https://github.com/inbo/etn/blob/solve-animal-tag-relationship/tests/testthat/test-get_animals.R#L163-L172

I can use basic R `grepl` function instead of `str_detect` from `stringr` package to remove a dependency. And maybe grepl can also avoid using  rlang function `!!`? I can test it now.

damianooldoni:
See https://github.com/inbo/etn/pull/106/commits/071764074a9d5ac44123bacf7137f8ffa7923022. I replaced functions of stringr, purr and rlang from `get_animals` and `test-get_animals`. 
The package `rlang` is however needed to load `.data`, which is in use in many pipes with dplyr functions (e.g. `df %>% filter(.data, condition)`). Using `.data` we don't get tens of notes while checking the package.

I think everything is now ok. Package remains very light, code is even more compact and functionality is added as requested.

R/zzz.R
```diff
@@ -115,6 +115,8 @@ collapse_transformer <- function(regex = "[*]$", ...) {
 #'
 #' @keywords internal
 #'
+#' @noRd
```
peterdesmet: @damianooldoni Aha! Good to know.

Issue #121, opened 2020-08-25
Default "order by"


peterdesmet:
Currently, data are returned as they are in the view. This is not always the most convenient for users to explore. Should they be returned with a default sort? @IPauwels @jreubens @PieterjanVerhelst 

Conclusion:

- [x] animals: by `animal_project_code`, `release_date_time` and then `tag_id`. Same is used by Movebank
- [x] tags: by `tag_id`
- [x] detections: by `tag_id` and then `date_time`. Same is used by Movebank
- [x] deployments: `network_project_code`, `station_name`, `deploy_date_time`
- [x] receivers: by `receiver_id`
- [x] projects: by `project_code`

IPauwels:
This would be my suggestion:
- animals by release_date_time
- tags by tag_id 
- detections by date_time
- deployments by station_name?
- receivers by receiver_id

jreubens:
I agree with Ine

peterdesmet:
There is also `get_projects()`. Should this be by `project_code` or `start_date` (not always populated)

IPauwels:
- Detections: that is fine by me and a good idea
- Deployments: I think I would like it as a user to have the deployments sorted as how they are geographically located, but I understand that consistency is also important, so then I would prefer perhaps indeed receiver_id. It is up to the scientist then to account for the fact that a minority of them might have changed location during the project.

Regarding get_projects(), I would say project_code. 

I think for a lot of cases Movebank is a good reference.

jreubens:
For get_projects, I agree with project_code.
For detections I agree as well
For Deployments however, it occurs that several receivers are used at one station (in different deployments).
Sorting by station_name and then on date_time seems more logic to me. But I guess it won't differ too much...

peterdesmet:
That's two votes for sorting by station_name, which makes sense, as it is geographical.

1. One issue is a bit that `station_names` are not very controlled, so sorting by `station_name` and `deploy_date_time` mixes projects:

pk | receiver_id | network_project_code | station_name | station_description | deploy_date_time
--- | --- | --- | --- | --- | ---
5552 | VR2-6481 | EMMN | 12 | 12 | 2007-04-25T00:00:00Z
4274 | VR2W-102239 | Artevigo | 12 | vinhos 12 | 2011-03-01T00:00:00Z
4194 | VR2W-119113 | OTN-Tosenfjorden | 12 | Tosenfjorden | 2015-04-07T08:00:00Z
[...]

PR #123, opened 2020-08-27
Order returned results

peterdesmet:
Addresses #121


damianooldoni (approved):
THanks @peterdesmet for this nice work. Everything works as wished.

Issue #130, opened 2020-10-07
Create function to download data for publication


peterdesmet:
Create function that replicates functionality of https://github.com/inbo/etn-occurrences/blob/master/src/1_download.Rmd

`download_dataset()`

## Input

- `directory`: where files should be put
- `animal_project_code`
- `scientific_name` (to filter e.g. `sync tag` out)
- ~~`cut_off_date`: is this necessary? **see below: at least not for deployments**. Undecided for detections~~

## Output

- `animals.csv`: associated with that `animal_project_code`
- `tags.csv`: associated with those animals
- `detections.csv`: associated with that `animal_project_code`. Also extract `network_project_codes`
- `deployments.csv`: all associated with set of `network_project_codes`
- `receivers.csv`: all found in deployments
- `datapackage.json`: metadata file describing those records
- A summary of the export in the terminal

peterdesmet:
@PieterjanVerhelst (in conversation in Google spreadsheet) says cut-off date for deployments not necessary:

> Een einddatum geeft goed weer wanneer een netwerk de laatste keer werd uitgelezen en is bijgevolg een indicatie voor zijn 'activity', dus zou dit zo laten.

peterdesmet:
@jreubens @PieterjanVerhelst will it ever be useful to download sync/sentinel data for an animal project as a dataset?

peterdesmet:
This is now implemented in #132, see online function documentation at https://inbo.github.io/etn/reference/download_dataset.html.

@jreubens @PieterjanVerhelst The above question regarding sync/sentinel data remains. If a scientific_name is not provided, the function will download all data including sync/sentinel data. This can be a lot (for a default option).

PieterjanVerhelst:
@jreubens we think it would be better to put sync/sentinel/range_test tags and data under a separate project. This would make those 'sub' datasets better traceable. For instance, researchers may be interested in a range test rather than the fish detections. Putting them in another project will make their existence clear.

jreubens:
@PieterjanVerhelst; @peterdesmet  I wouldn't put this type of data under a separate project.
It is often part of the analyses you perform in the framework of you species data as it is a proxy for environmental factors influencing your detectability...
However, I do agree that you are not always interested in this part of the data and perform the analysis separate from your species analysis. Can you add 'tag_type' in the function? THis is a compulsory field if you create a new tag and specifies if it is an animal, sync, built-in,.... thus than you can easily filter.

peterdesmet:
[...]

PR #132, opened 2020-10-13
Create function to download dataset

peterdesmet:
Adds a function `download_dataset()`, that bundles all the functionality of this script: https://github.com/inbo/etn-occurrences/blob/master/src/1_download.Rmd. See #130

@damianooldoni care to review/test? `2014_demer` is a good one.


R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
+#' Download data for publication
+#'
+#' ...
```
damianooldoni, line 3: Could you maybe write one or two sentences about it?
Just an idea: _This function allows to download data as csv files. The data are downloaded as they are in database, i.e. no consistency check is performed. We recommend to check them by .. before publication._

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
...
+#'
+#' @param connection A valid connection to the ETN database.
+#' @param animal_project_code (string) Animal project to download data from.
+#' @param directory (string) Path to local download directory.
+#' @param scientific_name (string) One or more scientific names to filter on.
+#'
+#' @return ...
```
damianooldoni, line 10: This should be surely be present. 
Something like:
A list of csv files:
1. `filename1.csv`
2. ...

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
...
+#'
+#' @export
+#'
+#' @importFrom glue glue
+#' @importFrom dplyr %>% arrange distinct filter group_by mutate pull select tibble
+#'
+#' @examples
+#' \dontrun{
+#' ...
```
damianooldoni, line 19: I can do this as I am planning to write some tests and a subset of the tests can perfectly be reused in this section.

damianooldoni, line 19: I can do this as I am planning to write some tests and a subset of the tests can perfectly be reused in this section.

R/download_dataset.R
peterdesmet: Yes, that part wasn't finished yet indeed!

R/download_dataset.R
peterdesmet: Agree

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
+#' Download data for publication
+#'
+#' ...
+#'
+#' @param connection A valid connection to the ETN database.
+#' @param animal_project_code (string) Animal project to download data from.
+#' @param directory (string) Path to local download directory.
```
damianooldoni, line 7: I would change "Path" with "Relative path"

damianooldoni: And also mentioning the default value.

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
+#' Download data for publication
+#'
+#' ...
+#'
+#' @param connection A valid connection to the ETN database.
+#' @param animal_project_code (string) Animal project to download data from.
```
damianooldoni, line 6: It's common practice to add default value while documenting args. It seems we have never done before, but I think this is a good function to start with. Especially in this function is quite important because if no animal project is assigned (`animal_project_code = NULL`), ALL data are download, and that's quite a big difference in terms of file dimensions, I think.

I would add the following:
Default: `NULL`. If `NULL`, the data from all animal projects in etn database are downloaded.

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
...
+#' @param directory (string) Path to local download directory.
+#' @param scientific_name (string) One or more scientific names to filter on.
+#'
+#' @return ...
+#'
+#' @export
+#'
+#' @importFrom glue glue
```
damianooldoni, line 14: We should also add:

```r
#' @importFrom readr write_csv
```

R/download_dataset.R
peterdesmet: Now documented

R/download_dataset.R
peterdesmet: I would not download all data by default! animal_project_code should be a required parameter, not sure how to do that consistently. Probably by not adding `= NULL`?

R/download_dataset.R
peterdesmet: This has now been updated + default is mentioned.

R/download_dataset.R
peterdesmet: Now added

R/download_dataset.R
damianooldoni: Yes, indeed. REmoving `= NULL` you make that argument mandatory for user.

peterdesmet: I have changed this to not adding `=NULL` with a check on `is.null()` and `missing()`. It is now also documented what the defaults are, but only for this function.

peterdesmet: See e7cb43b93e09a37fa0c879e51cd915e6e5dedbdd I have now done this for all functions.

R/download_dataset.R
damianooldoni: Confirmed. Thanks. You were faster than me.

R/download_dataset.R
```diff
@@ -0,0 +1,170 @@
...
+    stop("Please provide an animal_project_code")
+  } else {
+    valid_animal_project_codes <- list_animal_project_codes(connection)
+    check_value(animal_project_code, valid_animal_project_codes, "animal_project_code")
+  }
+
+  # Check directory
+  if (!dir.exists(directory)) {
+    dir.create(directory)
+  }
```
damianooldoni, line 89: Based on a comment of Thierry while developing `setup_codingclub_session()` I learned that we can avoid the if. See https://github.com/inbo/inborutils/blob/master/R/setup_codingclub_session.R#L129

damianooldoni (approved):
Nice function and well written. I wrote some single comments here and there as you could see yesterday.

About testing the function, I will write some tests but if you would like to use the function already or you know somebody who has to use it urgently, you can surely merge all to master.

I have only one "enhancement" question: when I think about data packages I directly don't think about a folder but a "package", a container, an archive, i.e. something like a zipped file. Shouldn't we could provide the option of making a zipped file instead of producing csv files? If this is in contrast with the guidelines of frictionless data, leave it. It's just an idea.

R/download_dataset.R
peterdesmet: Resolved

peterdesmet:
> Shouldn't we could provide the option of making a zipped file instead of producing csv files? If this is in contrast with the guidelines of frictionless data, leave it. It's just an idea.

Good suggestion, but I think we should leave it unzipped, because 1) we use the parameter `directory` 2) we want people to quality check the data (so they would have to unzip) and 3) files need to be separately to a repository if you want to allow them to be read by the Frictionless Framework

damianooldoni:
I found also that if same directory is given, files are overwritten without any warning, or so. Personally I find it quite dangerous, as somebody could make mistakes by overwriting files of different projects if they forget to change directory or so. I think a warning() is sufficient saying that "file x is present and will be overwritten"

damianooldoni:
We agree that a general additional message is sufficient. :+1: 
I would also inform people in function documentation.

Issue #141, opened 2020-11-18
Avoid arising warning in summary for empty projects in download_dataset()


damianooldoni:
While trying to download data of a project, `2018_Iller`, I get an additional warning at the end:

```r
Warning messages:
1: In min.default(numeric(0), na.rm = FALSE) :
  no non-missing arguments to min; returning Inf
2: In max.default(numeric(0), na.rm = FALSE) :
  no non-missing arguments to max; returning -Inf
```

This is due to the fact that the project is empty, so no min or max date for `detections` can be returned:

```
message("* first date of detection:     ", detections %>% summarize(min(as.Date(.data$date_time))) %>% pull())
message("* last date of detection:      ", detections %>% summarize(max(as.Date(.data$date_time))) %>% pull())
````

I would fix it by adding a small `if (nrow(detections) == 0) {}- else{}` to embed these two lines of codes.

@peterdesmet: is it ok for you?

peterdesmet:
Yes, good catch.

damianooldoni:
Fixed in https://github.com/inbo/etn/commit/1bace2b513c3fd007e834a9a970113764e9ca9a7

damianooldoni:
@peterdesmet, this issue has been solved. We can close it, isn't?

peterdesmet:
Yes indeed.

Issue #139, opened 2020-11-18
Documentation of download_dataset(): line width wider than 100 characters


damianooldoni:
While running devtools::check(), one of the notes I get is the following:

```r
checking Rd line widths ... NOTE
  Rd file 'download_dataset.Rd':
    \examples lines wider than 100 characters:
       #> * included scientific names:   Petromyzon marinus, Rutilus rutilus, Silurus glanis, Squalius cephalus
```

@peterdesmet: Is there a reason to use the animal project `2014_demer` in `@example`? Using a smaller project (aka less scientific names) would solve this note easily, e.g. `2015_homarus`.  

I can do it, I will include it in the PR I am writing now.

peterdesmet:
I'm using `2014_demer` because it showcases some of the warning messages

damianooldoni:
I understand. Good point. :+1:  
What about `2012_leopoldkanaal` then? It showcases a warning message and satisfies the check.

```r
#' # Download data for the 2012_leopoldkanaal animal project (all scientific names)
#' download_dataset(con, animal_project_code = "2012_leopoldkanaal")
#' #> Downloading data to directory "2012_leopoldkanaal":
#' #> (existing files with the same name will be overwritten)
#' #> * (1/6): downloading animals.csv
#' #> * (2/6): downloading tags.csv
#' #> * (3/6): downloading detections.csv
#' #> * (4/6): downloading deployments.csv
#' #> * (5/6): downloading receivers.csv
#' #> * (6/6): adding datapackage.json as file metadata

#' #> Summary statistics for dataset "2012_leopoldkanaal":
#' #> * number of animals:           104
#' #> * number of tags:              103
#' #> * number of detections:        2215839
#' #> * number of deployments:       1581
#' #> * number of receivers:         273
#' #> * first date of detection:     2012-07-03
#' #> * last date of detection:      2020-09-28
#' #> * included scientific names:   Anguilla anguilla
#' #> * included network projects:   albert, bpns, JJ_Belwind, leopold, pc4c, ws2, zeeschelde

#' #> Warning message:
#' #> In download_dataset(con, animal_project_code = "2012_leopoldkanaal") :
#' #> Found tags associated with multiple animals: A69-1601-29925
```

peterdesmet:
Ok for me!

Issue #138, opened 2020-11-18
Bug with connection arg in download_dataset()


damianooldoni:
This doesn't work:

```r
con1 <- connect_to_etn(
  username = Sys.getenv("userid"),
  password = Sys.getenv("pwd")
)
download_dataset(connection = con1, animal_project_code = "2014_demer")
Error in is(connection, "PostgreSQL") : object 'con' not found
```

The reason is that the get_*() functions called in `download_dataset()` are using `con` instead of `connection`. 
This bug is hidden in some way if you use default `con` instead of another variable name as R uses any `con` it can find in global environment.

Bug found thanks to unit-tests by the way :test_tube: even if at first I thought it was a problem while using `testthat::evaluate_promise()`.
 
@peterdesmet: I am solving it in branch `add-test-download_dataset` as PR is coming soon.

PR #142, opened 2020-11-19
Add test download dataset

damianooldoni:
This PR does the following:

1. Add unit-tests for `donwload_dataset()`
2. Add `readr` to `Imports` in `DESCRIPTION` as the readr's, function `write_csv()` is used in `download_dataset()`
2. Add check of `limit` as logical in `get_detections()`
3. ~Refactoring, updating and in some cases simplify tests of `get_animals()` and `get_detections()` as not all of them passed~
4. Replace `%\VignetteIndexEntry{Vignette Title}` in the vignettes header should be changed to the real title as mentioned in https://bookdown.org/yihui/rmarkdown-cookbook/package-vignette.html This is needed as we have now two vignettes, otherwise error arises
5. Replace `library(tidyverse)` by explicitly naming the needed packages in vignette `etn_fields.Rmd`  and add them in `Suggests` section of file `DESCRIPTION`, if they are not already imported or suggested
6. Solve other minor problems which come out while running devtools::check(), e.g. the typical note for not finding global variables which is partially solved as always by adding `.data$` to column names in dplyr functions
7. Fix bug described in #138 
8. Fix enhancement issues #139 and #141


peterdesmet (changes requested):
The structure of the tests for `get_detections()` and `get_animals()` are now very different from the other functions. Could you leave them unchanged in this PR, and submit those as another PR?

damianooldoni:
I changed those tests based on what I am learning from the book about unit-testing I am reading in my freetime :nerd_face: 

Yes, making the changes in a separate PR is ok to me.

peterdesmet (changes requested):
Also, `2015_homarus_detections_stubbed.csv` adds 4MB of data. I would suggest to use a zipped version of this file, or even better, the whole dataset: it will be deposited soon. I hope unzipping during the test is possible (and time efficient).

damianooldoni:
It seems also that number of receivers and deployments has changed yesterday. I had to update the csv containing the expected output (see last commit https://github.com/inbo/etn/pull/142/commits/c2ec590e7a89c48e8aadc351a4aa89a3b03292d1).

This makes me think that this project is still open?

damianooldoni:
@peterdesmet: I undid the changes of the unit-tests of `get_animals()` and `get_detections()`. Please, be aware that some of them will not pass now. As you suggested, I will start now a new branch for the improvement of these tests and I will make immediately after a new PR.

damianooldoni:
About working with zip file instead of csv file, I found it is not worth working with stubbing. With a slight faster internet connection is getting detections faster than reading zip/csv files. So, I removed `mockery` and `zip` from `Suggests` in `DESCRIPTION` as not anymore needed.

peterdesmet:
Great! No need for an included test file, just using the `download_dataset()` function.

- I have moved the message output example
- I have updated `get_animals()` and `get_detections()`test projects so they pass (will create conflict, but don't want to leave it in this state)

```
0 errors ✓ | 1 warning x | 1 note x
```

@damianooldoni you can merge this.

peterdesmet (approved):

PR #154, opened 2021-05-16
Restore functionality with old database

peterdesmet:


peterdesmet:
The pkg is now functional again (by using `acoustic` rather than `vliz` database schema). The vignette `etn_fields` had to be updated as well, since the old `field_definitions` table no longer exists. This caused a number of changes (#155 and #156) that need to be fixed later.

The pkg is using a legacy copy of the database. I added a warning message on the homepage to clarify this.

PR #241, opened 2021-11-29
Fix #240: use :: over importFrom

### LINKED ISSUE #240: Don't use @importFrom (except for %>%), but call with ::
See https://github.com/frictionlessdata/frictionless-r/commit/750eef30fa1e29d5bd5611e99ae1ef748c90de95

Makes for clearer code (i.e. where does a function come from).

peterdesmet:
Don for all functions, except for `dplyr %>% .data distinct filter pull`, which are imported in `etn.R`


damianooldoni (approved):
Thanks for doing this! Approved.

PR #253, opened 2022-09-14
Fix tests

peterdesmet:
Fix tests. Issues are created for tests that have been disabled.


damianooldoni (approved):
Nice work, @peterdesmet. It's ok to comment such tests while issue is still open.

Issue #256, opened 2022-11-21
Export data as Darwin Core


peterdesmet:
Add function to export data as Darwin Core

## Human Observations

- [x] type
- [x] license: provided by user, CC-BY / CC0
- [x] rightsHolder: provided by user
- [x] datasetID: URL including imis_dataset_id
- [x] institutionCode: VLIZ
- [x] collectionCode: ETN
- [x] datasetName: imis title
- [x] basisOfRecord
- [x] dataGeneralizations: NULL
- [x] occurrenceID
- [x] sex
- [x] lifeStage
- [x] occurrenceStatus
- [x] organismID
- [x] organismName
- [x] eventID
- [x] parentEventID
- [x] eventDate
- [x] samplingProtocol
- [x] samplingEffort
- [x] eventRemarks
- [x] locationID: NULL
- [x] locality
- [x] decimalLatitude
- [x] decimalLongitude
- [x] geodeticDatum
- [x] coordinateUncertaintyInMeters: 30
- [x] scientificNameID
- [x] scientificName
- [x] kingdom: Animalia

## Detections

- [x] type
- [x] license
- [x] rightsHolder
- [x] datasetID
- [x] institutionCode
- [x] collectionCode
- [x] datasetName
- [x] basisOfRecord
- [x] dataGeneralizations: `subsampled by hour: first of 3 record(s)	`
- [x] occurrenceID: detection id_pk
- [x] sex: continued from deployment (cf. Movebank)
- [x] lifeStage: not set, as it can change through lifetime
- [x] occurrenceStatus
- [x] organismID
- [x] organismName
- [x] eventID
- [x] parentEventID
- [x] eventDate
- [x] samplingProtocol: `acoustic telemetry`
- [x] eventRemarks: NULL
- [x] locationID: deployment_station_name
- [x] locality: location_name
- [x] decimalLatitude
- [x] decimalLongitude
- [x] geodeticDatum
- [x] coordinateUncertaintyInMeters: set to 1000m
- [x] scientificNameID
- [x] scientificName
- [x] kingdom

peterdesmet:
@jreubens @jonasmortelmansvliz is there a standardized value to be used for license? Is it always CC0 (recommended) or CC-BY?

peterdesmet:
I notice that the `coordinating_organization` as stored in the DB can be really long:

> Vlaamse overheid; Beleidsdomein Omgeving; Instituut voor Natuur- en Bosonderzoek

vs

> INBO

[...]

PR #257, opened 2022-11-23
Add `write_dwc()` function

peterdesmet:
Implements #256

@jreubens @jonasmortelmansvliz @damianooldoni @jdpye @sarahcd here's a first version of the `write_dwc()` function that transforms acoustic telemetry data to Darwin Core that can be harvested by OBIS and GBIF. It follows the recommendations we discussed, which are based on conventions introduced for Movebank data in the [movepub `write_dwc()` function](https://inbo.github.io/movepub/reference/write_dwc.html). An `eml.xml` file is not created.

You can test it by installing:

```R
devtools::install_github("inbo/etn#257")
```

@jdpye @sarahcd this will work on LifeWatch RStudio Server only for now, but you can see the SQL file in this PR to see the transformation.

Then:

```R
library(etn)
con <- connect_to_etn()
write_dwc(
  animal_project_code = "2014_demer" # or other project
  # optional parameters are: directory, rights_holder, license
```

If all goes well, a `dwc_occurrence.csv` will be written to disk.


peterdesmet:
Note: `dataGeneralizations` currently has the fixed value `subsampled by hour`. The goal is to make this `subsampled by hour: first of 3 record(s)` cf. Movebank. Calculating the number of records by group is however a more costly query than the current approach. Do you consider it worth it to provide this information?

peterdesmet:
For those who cannot run the function, here's the DwC file [dwc_occurrence.csv](https://github.com/inbo/etn/files/10112135/dwc_occurrence.csv) for the following query:

```
write_dwc(animal_project_code = "2014_demer", rights_holder = "INBO")
```

peterdesmet:
Regarding https://github.com/inbo/etn/pull/257#issuecomment-1324900156:

Count per group is now included as `subsampled by hour: first of 6 record(s)` (cf. Movebank). Duplicates are excluded. The function takes the same amount of time as previously (35 seconds for a big project like `2013_albertkanaal `).

peterdesmet:
Only a couple of pending questions to finalize this: https://github.com/inbo/etn/issues/256#issuecomment-1330485004

peterdesmet:
All Darwin Core mapping issues are resolved (#256).

@jreubens @jonasmortelmansvliz @jdpye note that the current implementation only exports an **Occurrence core** (cf. Movebank data), not an **Extended measurement or fact extension**. The latter could be added in the future, but it should then be discussed what to express in there (scope potentially huge).

This PR is pending technical review by @PietrH.

R/write_dwc.R
```diff
@@ -0,0 +1,99 @@
...
+#' - Acoustic detections are downsampled to the **first detection per hour**,
+#'   to reduce the size of high-frequency data.
+#'   Duplicate detections (same animal, tag and timestamp) are excluded.
+#'   It is possible for a deployment to contain no detections, e.g. if the
+#'   tag malfunctioned right after deployment.
+write_dwc <- function(connection = con,
+                      animal_project_code,
+                      directory = ".",
+                      rights_holder = NULL,
+                      license = "CC0") {
```
PietrH, line 47: I believe this should be CC-BY as default, see [this comment](https://github.com/inbo/etn/issues/256#issuecomment-1323664526) on #256

inst/sql/dwc_occurrence.sql
```diff
@@ -0,0 +1,252 @@
...
+  NULL                                  AS locationID,
+  event.locality                        AS locality,
+  event.latitude                        AS decimalLatitude,
+  event.longitude                       AS decimalLongitude,
+  CASE
+    WHEN event.latitude IS NOT NULL THEN 'EPSG:4326'
+  END                                   AS geodeticDatum,
+  CASE
+     -- Assume coordinate precision of 0.001 degree (157m) and recording by GPS (30m)
+    WHEN event.latitude IS NOT NULL THEN 187
```
PietrH, line 178: `WHEN event.latitude IS NOT NULL THEN 187` should be 157 as per comment on line 177

PietrH: 157 for rounding + 30 for GPS

R/write_dwc.R
jreubens: We indeed recommand CC-by for all ETN datasets

peterdesmet: Ok, I can change to CC-BY even though CC0 makes more sense for scientific data. 🙂

PietrH (changes requested):
Before merging I suggest the following changes:

- [x] replace expect_equal with expect_identical where appropriate
- [ ] Allow write_dwc() to return objects rather than files

The latter similar to inbo/camtraptor#181

tests/testthat/test-check_value.R
```diff
@@ -1,52 +1,44 @@
...
+    check_value("invalid", c("a", "b"), name = "param_name"),
+    "Can't find param_name `invalid` in: a, b",
+    fixed = TRUE
   )
-  expect_true(
-    check_value(NULL, c("animal", "network"), "project_type")
+})
+
+test_that("check_value() returns x for correct values", {
+  expect_equal(
```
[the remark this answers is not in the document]
peterdesmet: Fixed for this code. I created a separate issue for the rest: #268

peterdesmet:
@PietrH all comments addressed.

PietrH (approved):
Looks good to me.

sarahcd:
Super belatedly, I checked out the sql and example output dwc_occurrence (without access to the LifeWatch server), it all looks sensible to me. Happy to coordinate with ETN at any point on aligning future changes with Movebank. Thank you for all the work @peterdesmet !

Issue #242, opened 2022-03-08
OpenCPU approach


peterdesmet:
Summary of March 8 meeting:

Create two "flavours" of all functions in the etn package: **local** and **remote**. Function names remain the same, the difference is made in the `con` variable. For local `con` is a database connection, for remote `con` are credentials to be passed via OpenCPU. The intent is to keep a single R package.
- Users on rstudio.lifewatch.be use the local flavour (and connect directly to database)
- Package on Docker will use the local flavour to connect to database. All functions will be exposed as OpenCPU endpoints (unless the access is closed via Apache).
- Anyone can use the remote flavour and connect via OpenCPU.

For remote access, bandwidth and file size might become an issue. Potential solutions:
- ~Transfer data as binary [feather](https://www.opencpu.org/api.html#api-ioformats) files. Read with `arrow::read_feather()`.~ Feather files are not compressed, so there is no gain here over csv. Most compressed format is very likely`rda`, so planning to use that.
- Use paging
- Only offer `download_acoustic_detections()` (rather than `get_accoustic_detections()`) so users only have to transfer huge amounts of data once.

Next steps:
- [x] Get etn package working on docker, including DB connection. by @salvafern
- [x] Make some of the OpenCPU endpoints public. by @salvafern
- [ ] Test endpoints to design remote flavour of functions. by @peterdesmet
- [ ] Update package to make it work fully remotely. by @peterdesmet

salvafern:
I changed slightly one function to make a connection to the etn directly by providing your username and password in https://github.com/inbo/etn/commit/34254e2db2c8c048094e1566d6f4a41582f1cbe3.

Now it works fine with opencpu! so it seems that we have to go for this solution instead of using the connection object. I hope this is ok for you @peterdesmet?

peterdesmet:
Great! For backwards compatibility, I suggest to keep using the `con` variable as a single parameter, e.g. as a list with:

```R
con <- list(
  username = "x",
  password = "y"
)
```

That avoids parameters to be shifted in functions:

```R
get_animals(my_con, 305)
# still calls:
get_animals(con = my_con, animal_id = 305)
# rather than:
get_animals(username = my_con, password = 305)
```

salvafern:
[...]

PR #271, opened 2022-12-20
Opencpu client dev

PietrH:
# STATUS: NOT READY FOR REVIEW

Introducing a list_animal_ids() version that uses the OpenCPU API instead of local database access: list_animal_ids_api(), this function currently uses the optional json flag in the POST request and thus has to parse JSON to return a vector. 

## Dependencies
This pull request introduces a new dependency on [httr](https://httr.r-lib.org/), a wrapper of curl. 

## New helpers
- get_credentials() fetches user credentials from the system environment, needed to access the api. These are then passed directly as strings in the HTTP request body. (needs review)
- extract_temp_key() extracts the temp key from the path returned by a non-json POST request to the API.
- get_val() takes the response object of a POST request to the api, and places a GET request for the evaluated response of the function. Uses RDS as an intermediary for compression.

The last two helpers are currently unused. I'm leaving these in for the moment in case we want to test compression for retrieval of bigger objects later on.

## Branch Management
I'm planning to keep development on separate branches, and merge to OpenCPU-client, keeping it as up to date with main as possible. 

## Tests
Currently, the tests check the result of the API, not the API performance or if it's reachable (although if you can't reach it, you'll still fail the test). 

## What's next
- Add check for get_credentials that credentials are set in the environment, instead of just returning empty strings. 
related to #242
- Deploy etnservice (new API compliant version of `get_acoustic_detections()`)
- Develop `get_acoustic_detections_api()`
- Select testing framework for the API (medium term)
- Setup tests of API for uptime/performance (medium term)

## TODO
following discussion in #280: I'm still using `get_acoustic_detections()` as a testcase. 

- [x] API: better error handling in `get_credentials()`
- [x] use helpers for api or sql handling
- [x] remove `get_acoustic_detections_api()` from exports
- [ ] write tests for ~~api route~~ sql route
- [x] wrap API code in generic helper
- [x] wrap main function code: `if(api){out<-api_out}else{out<-sql_out} return(out)` in generic wrapper
- [x] have helpers output to object that is handled in main function body
- [x] update `get_acoustic_detections()` documentation to reflect #280
- [ ] update `list_animal_ids()` documentation
- [x] double check api fetching method is rda
- [x] adapt `list_animal_ids()` to use the helper structure from #280
- [x] pass error messages and warnings from api to client
- [x] set rate limiting for first step of api http request via helper OR use `httr::RETRY()` [ref](https://httr.r-lib.org/reference/RETRY.html)
- [x] set rate limiting for helper: `get_val()`
- [x] rewrite `list_acoustic_project_codes()`
- [x] rewrite `list_acoustic_tag_ids()`
- [x] rewrite `list_animal_project_codes()`
- [x] rewrite `list_receiver_ids()`
- [x] rewrite `list_scientific_names()`
- [x] rewrite `list_station_names()`
- [x] add assertation for `api` argument
- [x] close connections before returning output in _sql helpers
- [ ] split off helpers from utils.R
- [ ] 
### Testing effort

- [ ] add tests for api helpers
- [ ] achieve full coverage for api helpers
- [ ] add tests for sql flow
- [ ] achieve full coverage for api flow
- [ ] skip sql tests when in CI

### Documentation

- [ ] create documentation for new `api` argument
- [ ] Have functions inherit `api` argument documentation
- [ ] add deprecation badge to `connection` argument
- [ ] add deprecation documentation to `connection` argument
- [ ] Have functions inherit `connection` argument documentation
- [ ] port examples over to use api


PietrH:
We should probably merge main into opencpu-client as well

peterdesmet (changes requested):
- [ ] Although the function can be named `list_animal_ids_api()` during development of v3, I would include it in `list_animal_ids.R` not in a separate file. I think that will make for a cleaner version history and will be easier to review params etc.
- [ ] Place `get_credentials()` in a separate `get_credentials.R` file. `utils.R` should be reserved for internal functions
- [ ] Rename `opencpu-client` branch to `v3` as discussed
- [ ] You mention functions currently not used. I would list those in a todo issue, so they don't linger in the code

R/list_animal_ids_api.R
```diff
@@ -0,0 +1,27 @@
+#' List all available animal ids
+#'
+#' @param credentials Login credentials to the ETN database, as created by
+#'   `get_credentials()`
+#'
+#' @return A vector of all unique `id_pk` present in `common.animal_release`.
+#'
+#' @export
+list_animal_ids_api <- function(credentials = get_credentials()) {
+  api_domain <- "https://opencpu.lifewatch.be"
```
peterdesmet, line 10: I find the split between `api_domain` and `function_path` a bit arbitrary (parts are static in both) and would just combine them into one variable `endpoint` (I see you called it `api_url` in an example below.):
```R
endpoint <- "https://opencpu.lifewatch.be/library/etn/R/list_animal_ids"
```

R/list_animal_ids_api.R
```diff
@@ -0,0 +1,27 @@
...
+  function_path <- "library/etn/R/list_animal_ids"
+
+  # POST the function request to the api_domain, request json as direct return
+  response <-
+    httr::POST(stringr::str_glue(api_domain, function_path, "json", .sep = "/"),
```
peterdesmet, line 15: Please use `glue::glue()` over `stringr::str_glue()`. We use glue directly in our packages, as it is very lightweight (no dependencies), while stringr is a more of a conscious dependency.

peterdesmet: Note that here we could also use `paste()`. I typically use `paste()` for simple things like this and`glue()` when I want to insert variables.

R/list_animal_ids_api.R
```diff
@@ -0,0 +1,27 @@
...
+      body = list(
+        con = credentials
+      )
+    )
+  # If request was not succesful, generate a warning
```
peterdesmet, line 20: Typo `successful`

R/list_animal_ids_api.R
```diff
@@ -0,0 +1,27 @@
...
+  httr::warn_for_status(response, "submit request to API server")
```
peterdesmet, line 21: When will this warn?

- `https://opencpu.lifewatch.com/library/etn/R/list_animal_ids/json` (.com) returns error when calling
- `https://opencpu.lifewatch.be/library/etn/R/list_animal_ids/js` (js) doesn't warn

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
...
+#'
+#' @param username ETN Data username, by default read from the environment, but
+#'   you can set it manually too.
+#' @param password ETN Data password, by default read from the environment, but
+#'   you can set it manually too.
+#'
+#' @return A string as it is ingested by other functions that need
+#'   authentication
+#' @family helper functions
+#' @noRd
```
peterdesmet, line 105: Since this function is referenced in the documentation of params, I would make this a public function (even if that means users can store their credentials in a variable). I notice the same is done for e.g. `show_progress()`, `readr_threads()`  in https://readr.tidyverse.org/reference/read_delim.html

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
   )
   as.character(parsed)
 }
+
+#' Get the credentials from environment variables, or set them manually
+#'
+#' By default, it's not necessary to set any values in this function as it's
+#' used in the background by other functions. However, if you wish to provide
```
peterdesmet, line 93: For clearer versioning, start every new sentence on a new line.

```
#' used in the background by other functions.
#' However, if you wish to provide your username and password on a per function 
#' basis ...
```

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
   )
   as.character(parsed)
 }
+
+#' Get the credentials from environment variables, or set them manually
```
peterdesmet, line 90: Would make more succinct and refrain from using comma's in function titles:

```
Get login credentials
```

With the different credentials floating around (R Studio server, applications, DB) we should probably figure out how these can be called consistently.

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
...
+#' @param password ETN Data password, by default read from the environment, but
+#'   you can set it manually too.
+#'
+#' @return A string as it is ingested by other functions that need
+#'   authentication
+#' @family helper functions
+#' @noRd
+
+get_credentials <-
+  function(username = Sys.getenv("userid"),
```
peterdesmet, line 108: Don't start on a new line, use https://style.tidyverse.org/functions.html#long-lines-1

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
...
+           password = Sys.getenv("pwd")) {
+    stringr::str_glue('list(username = "{username}", password = "{password}")')
```
peterdesmet, line 110: Use `glue::glue()`

R/list_animal_ids_api.R
```diff
@@ -0,0 +1,27 @@
+#' List all available animal ids
+#'
+#' @param credentials Login credentials to the ETN database, as created by
+#'   `get_credentials()`
+#'
```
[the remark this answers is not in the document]
peterdesmet: Also fix for functions below

R/utils.R
peterdesmet: Also fix for functions below

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
...
+#' }
+#'
+#' # using the opencpu test instance
+#' api_url <- "https://cloud.opencpu.org/ocpu/library/stats/R/rnorm"
+#' httr::POST(api_url, body = list(n = 10, mean = 5)) %>%
+#'   extract_temp_key() %>%
+#'   get_val(api_domain = "https://cloud.opencpu.org/ocpu")
+get_val <- function(temp_key, api_domain = "https://opencpu.lifewatch.be") {
+  httr::GET(
+    stringr::str_glue(
```
peterdesmet, line 158: Use `glue::glue()`

R/utils.R
```diff
@@ -86,3 +86,83 @@ check_date_time <- function(date_time, date_name = "start_date") {
...
+#' @noRd
+extract_temp_key <- function(response) {
+  response %>%
+    httr::content(as = "text") %>%
+    stringr::str_extract("(?<=tmp\\/).{15}(?=\\/)")
+}
+
+#' Retrieve the result of a function called to the opencpu api
+#'
+#' Loading the evaluated object into the current environment, to be used
```
peterdesmet, line 136: I don't understand this long sentence. 😄

tests/testthat/test-list_animal_ids_api.R
```diff
@@ -0,0 +1,17 @@
+test_that("list_animal_ids_api() returns unique list of values", {
```
peterdesmet, line 1: I would not create a new test file for this, but include it in `test-list_animal_ids.R`.

peterdesmet (changes requested):

peterdesmet:
I'm wondering if the current approach of naming new functions `_api` is a good one? Wouldn't it be better to introduce the new functions with their final name (at the top of their respective file) and rename the old ones to e.g. `_db`  or `_sql` (pushed down in their current file until they are removed)?

The advantages I see are:
- The functions that will be _kept_ get their final name (no massive renaming at the end)
- Test suite for the current SQL functions can immediately be used for the new functions (they are still used for the sql functions in the main branch)
- No discrepancy between function name and file name (idem for test file names)

The only downside is that it isn't immediately clear which function now used the api and which one still used the DB, but I'm thinking an issue with a checklist could cover that?

R/list_animal_ids_api.R
PietrH: I'll combine them for now, but I might split them up later again. 

The code block where I create the response object is likely to be reused in other functions, thus might be wrapped in a helper with the body and function path (function name) as arguments. When I wrote this I was already experimenting with this approach, but decided not to go trough with it because it added unnecessary complexity at this stage (only one function actually using the API).

R/list_animal_ids_api.R
PietrH: It'll warn for 40x http response codes, since you passed an invalid content type as a post request (`js`), opencpu falls back to the POST -> GET protocol (instead of returning json) you get a string array, thus content-type text instead of application/json, but no warning since the call was succesful (HTTP 201)

We could add an extra test that checks if the server returns the right content type.

PietrH: I've added an extra test for this, but again, this should probably be wrapped in a helper function in the future as we want to check this for every request.

R/list_animal_ids_api.R
peterdesmet: Aha, I understand. A helper function would be good indeed.

R/list_animal_ids_api.R
peterdesmet: Yes, good to do this for the helper function.

PietrH: Done

R/utils.R
PietrH: ### Fetch the result of an API call to OpenCPU

This function is used internally to GET an evaluated object from an OpenCPU api, to GET a result, you must of course POST it first

PietrH:
Larger queries seem to be failing over the api, such as `get_acoustic_detections(acoustic_project_code = "demer", limit = FALSE, api = TRUE)`

PietrH:
Removed helper: `as_json_primitives()` because it's currently unused. I originally wrote it to pass R primitives directly to OpenCPU over a GET/POST request. However, httr has an internal translation layer that we can use instead. 

Just in case I still need it later, I've made a gist: 

https://gist.github.com/PietrH/2f15b66f0ff3e05a014338757f9449f3

PietrH:
I'm starting a larger development effort in this branch/PR to convert more functions to use the API. This is tracked in #281

PR #274, opened 2023-01-02
add depth_in_meters field to get_acoustic_detections() 

### LINKED ISSUE #261: Add depth in the detections output
There has been a new field added on the `acoustic.detections_limited` .
Is it possible to add `depth_in_meters` in the output for the detections?

### LINKED ISSUE #268: Replace `expect_equal()` with `expect_identical()` in tests where possible
##### Note from Pieter: 

reference: https://testthat.r-lib.org/reference/equality-expectations.html

> These functions provide two levels of strictness when comparing a computation to a reference value. expect_identical() is the baseline; **expect_equal() relaxes the test to ignore small numeric differences.**

> In the 3rd edition, both functions use [waldo](https://github.com/r-lib/waldo). They differ only in that expect_equal() sets tolerance = testthat_tolerance() so that small floating point differences are ignored; this also implies that (e.g.) 1 and 1L are treated as equal.

### LINKED ISSUE #275: Add test to validate generated data package
`download_acoustic_dataset()` creates 5 CSV files based on `get_` functions and adds a static [datapackage.json](https://github.com/inbo/etn/blob/main/inst/assets/datapackage.json) file, making it a Frictionless Data Package. Changes in the `get_` functions (e.g. adding a field, see #274) without updating `datapackage.json` would create invalid data packages.

It would therefore be good to have a test that verifies the output of `download_acoustic_dataset()` against its `datapackage.json`. That can be done using the frictionless R package, which can be added as a dev dependency.

The way I would do it is reading the datapackage and making sure no warnings occur when reading the resources.

```R
p <- frictionless::read_package("link/to/datapackage.json")
expect_no_warning(frictionless::read_resource(p, "animals"))
expect_no_warning(frictionless::read_resource(p, "tags"))
...
```

PietrH:
resolves #261
resolves #275 
resolves #268

## Main changes

- Added single line to sql query, adapted test to expect the new column
- Added missing to datapackage.json schema
- I also made some small stylistic changes to some other functions, removing trailing whitespace, reducing column width. 
- Bumped minor version number
- Created NEWS.md
- swapped `expect_equal()` to `expect_identical()` in tests, see #268

## Changes to tests

- The test for download_acoustic_dataset() now uses the OS temp dir instead of a folder in the package root.

- Add dependency on frictionless. In test-download_acoustic_dataset.R we now use frictionless to actually check if the produced datapackage can be read without any warnings. I've also added checks for any fields that might be missing from the schema or in the wrong order. This turned out to be the case, these were added to the datapackage.json file that is used for testing. 

- I've also switched over test-download_acoustic_dataset.R over to using snapshots to check for the download messaging instead of the boolean character matching from before. This often caused the test to fail for reasons that were unclear, this method stores a standardized markdown file that is committed of the messaging that can be examined via a diff. This workflow is part of the 3rd edition of testthat, and I've switched the package over to this version. This included making some changes to other tests, namely switching over `expect_is()` to `expect_type()`.

- The snapshot could also notify us if the result of demer_2014 changes, and this is how I implemented it originally, but at the moment I'm only checking the console out (cat). And not the actual files generated as these should be covered in the other tests. If we want to include this data in the package, we need to make sure the rights are cleared under the repo license. 

- I've silenced some of the messaging internal to the tests to clean up the console output during package checking. 

- Added a test for list_values to check the messaging

## Notes / Possible new issues

- [ ] The test for get_acoustic_detections() is slow
A number of optimizations are possible, most of the time goes to multi-select queries like `get_acoustic_detections(con, acoustic_tag_id = c("A69-1601-16129", "A69-1601-16130"))` that can take around ten seconds to complete. I was expecting queries like this to be quicker.

- [x] Quite a bit of code is more then 80 columns wide

We could consider opening issues for these. But neither is urgent since they only impact development workflow.

- [x] citation("etn") output is missing a year:

``` r
citation("etn")
#> Warning in citation("etn"): no date field in DESCRIPTION file of package 'etn'
#> Warning in citation("etn"): could not determine year for 'etn' from package
#> DESCRIPTION file
#> 
#> To cite package 'etn' in publications use:
#> 
#>   Peter Desmet, Damiano Oldoni and Stijn Van Hoey (NA). etn: Access
#>   Data from the European Tracking Network. https://github.com/inbo/etn,
#>   https://inbo.github.io/etn.
#> 
#> A BibTeX entry for LaTeX users is
#> 
#>   @Manual{,
#>     title = {etn: Access Data from the European Tracking Network},
#>     author = {Peter Desmet and Damiano Oldoni and Stijn {Van Hoey}},
#>     note = {https://github.com/inbo/etn, https://inbo.github.io/etn},
#>   }
```

<sup>Created on 2023-02-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Slightly different locally:

``` r
citation("etn")
#> Warning in citation("etn"): no date field in DESCRIPTION file of package 'etn'
#> Warning in citation("etn"): could not determine year for 'etn' from package
#> DESCRIPTION file
#> 
#> To cite package 'etn' in publications use:
#> 
#>   Desmet P, Oldoni D, Van Hoey S (????). _etn: Access Data from the
#>   European Tracking Network_. https://github.com/inbo/etn,
#>   https://inbo.github.io/etn.
#> 
#> A BibTeX entry for LaTeX users is
#> 
#>   @Manual{,
#>     title = {etn: Access Data from the European Tracking Network},
#>     author = {Peter Desmet and Damiano Oldoni and Stijn {Van Hoey}},
#>     note = {https://github.com/inbo/etn, https://inbo.github.io/etn},
#>   }
```

<sup>Created on 2023-02-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I suggest we add a year (year of latest release) for easy of copy pasting.


peterdesmet (approved):
Nice!

1. Any reasoning behind the position of the field?
2. Please also include the field in https://github.com/inbo/etn/blob/35e70f3d284cb64494313e667f238378564d494b/inst/assets/datapackage.json#L569-L572 (there is currently no test for this, will create issue for that)
3. Minor version bump sounds good to me

PietrH:
> Nice!
> 
>     1. Any reasoning behind the position of the field?
> 
>     2. Please also include the field in https://github.com/inbo/etn/blob/35e70f3d284cb64494313e667f238378564d494b/inst/assets/datapackage.json#L569-L572
>         (there is currently no test for this, will create issue for that)
> 
>     3. Minor version bump sounds good to me

## 1. Position of field
The documentation for get_acoustic_detections refers to the field definitions (which seem out of date), which refer to the following csv file: https://github.com/inbo/etn/blob/main/inst/assets/etn_fields.csv, in this csv on row 112, there is a field `sensor_value_depth`, which I suspect depth_in_meters is based on. I'm also reasoning that depth_in_meters is a property of the sensor, not the animal itself. Thus it should follow fields to do with the sensor rather than the animal. 

I don't have strong feelings about this, apart that I didn't want to add it to the end. I'm fine with moving it to be grouped with the animal fields such as scientific_name

## datapackage

- [ ] #275  
- [ ] Add depth in meters to datapackage.json 

## Minor version bump
Will do! Thank you for the review

peterdesmet:
Thanks, I would suggest to have the field immediately after `deploy_longitude` then, as part of the "location" information.

PietrH:
Will do

PietrH:
datapackage.json is in the wrong order, and is missing 5 more fields from the tags table: length, diameter, weight, floating,  and archive_memory. 

However, I can't find any examples of these fields in use in the tables I have access to, apart from archive memory in these 3 examples:

``` r
[1] "shad_scheldt_dst"
# A tibble: 1 × 5
  length diameter weight floating archive_memory
   <dbl>    <dbl>  <dbl> <chr>    <chr>         
1     NA       NA     NA NA       2 MB          
[2] "FISHINTEL"
# A tibble: 2 × 5
  length diameter weight floating archive_memory
   <dbl>    <dbl>  <dbl> <chr>    <chr>         
1     NA       NA     NA NA       2             
2     NA       NA     NA NA       8             
[3] "2018_EC"
# A tibble: 1 × 5
  length diameter weight floating archive_memory
   <dbl>    <dbl>  <dbl> <chr>    <chr>         
1     NA       NA     NA NA       64 MB  
```

It seems like archive_memory should be a string, length, diameter and weight seem to be doubles in the returned table, but I have no values to make sure. floating seems to be a string as well. 

This didn't come to light earlier because there was no test for this in test-download_acoustic_dataset.R

peterdesmet:
@PietrH these fields are directly taken from:

https://github.com/inbo/etn/blob/35e70f3d284cb64494313e667f238378564d494b/R/get_tags.R#L129-L133

The data type for those fields in the database is:

```R
DBI::dbGetQuery(con, "
  SELECT
    pg_typeof(archive_length) AS length,
    pg_typeof(archive_diameter) AS diameter,
    pg_typeof(archive_weight) AS weight,
    pg_typeof(archive_floating) AS floating,
    pg_typeof(device_internal_memory) AS memory
  FROM
    common.tag_device_limited
  LIMIT 1
")

            length         diameter           weight floating            memory
1 double precision double precision double precision  boolean character varying
```
Looks like dplyr interpreted the boolean as string 🤷‍♂️ 

So in frictionless:

- length: number
- diameter: number
- weight: number
- floating: boolean
- memory: string

PietrH:
Ready for final review, I've added notes to the first post of this pull request. Including some suggestions for new issues. 

If we are ready to merge, I'll create a new release. Do we have/want a NEWS.md?

peterdesmet:
We don't have a news, but you can start one as part of this PR. See https://github.com/frictionlessdata/frictionless-r/blob/HEAD/NEWS.md for inspiration.

PietrH:
I've added a NEWS.md to this PR, I've added selected changelogs based on the releases since 2.0.0 feel free to change this document directly to better reflect past changes.

R/get_acoustic_detections.R
```diff
@@ -199,7 +199,7 @@ get_acoustic_detections <- function(connection = con,
   # Check limit
   assertthat::assert_that(is.logical(limit), msg = "limit must be a logical: TRUE/FALSE.")
   if (limit) {
-    limit_query <- glue::glue_sql("LIMIT 100", .con = connection)
+    limit_query <- glue::glue_sql("ORDER BY det.id_pk LIMIT 100", .con = connection)
```
PietrH, line 202: Ordering is necessary to ensure the result of the filter is always the same, if we leave out the `ORDER BY`, then a test will fail. However, this step is very expensive, defeating the point of the `LIMIT`. 

Consider restoring to unordered limiting, but documenting the behavior and perhaps rewriting the test.

PietrH: Ordering makes the examples and testing stages too slow:

```md
✔  checking examples (40m 42.8s)
   Examples with CPU or elapsed time > 5s
                             user system  elapsed
   get_animals             12.828  0.092   15.457
   get_acoustic_detections 10.664  0.552 2406.708
   get_tags                 6.940  0.040    8.048
```

PietrH: Limiting actually seems to be a little bit slower than not limiting, at least in some circumstances: 


```md

Unit: seconds
                                                                              expr
  get_acoustic_detections(con, acoustic_project_code = "demer",      limit = TRUE)
 get_acoustic_detections(con, acoustic_project_code = "demer",      limit = FALSE)
      min       lq     mean   median       uq      max neval
 151.7327 154.1687 164.0103 156.6048 170.1491 183.6935     3
 110.5260 112.2894 123.5281 114.0527 130.0292 146.0056     3
```

tests/testthat/test-get_acoustic_detections.R
```diff
@@ -112,22 +117,27 @@ test_that("get_acoustic_detections() allows selecting on animal_project_code", {
...
-    get_acoustic_detections(con, animal_project_code = "2014_demer", limit = TRUE),
-    get_acoustic_detections(con, animal_project_code = "2014_DEMER", limit = TRUE)
+  expect_identical(
+    # a limited query doesn't always return the same results, the order isn't guaranteed
+    arrange(
+      get_acoustic_detections(con, animal_project_code = "2014_demer", limit = FALSE),
+      detection_id),
+    arrange(
+      get_acoustic_detections(con, animal_project_code = "2014_DEMER", limit = FALSE),
+      detection_id)
```
PietrH, line 134: These two (equivalent) queries each create a 33MB object, we should probably look for `animal_project_code` values that result in smaller objects for testing, that would speed things up.

tests/testthat/test-get_acoustic_detections.R
```diff
@@ -167,22 +177,25 @@ test_that("get_acoustic_detections() allows selecting on acoustic_project_code",
...
-  expect_equal(
+  expect_identical(
     single_select_df %>% distinct(acoustic_project_code) %>% pull(),
     c(single_select)
   )
   expect_gt(nrow(single_select_df), 0)
 
   # Selection is case insensitive
-  expect_equal(
-    get_acoustic_detections(con, acoustic_project_code = "demer", limit = TRUE),
```
PietrH, line 178: These two (equivalent) queries each create a 33MB object, we should probably look for `acoustic_project_code` values that result in smaller objects for testing, that would speed things up.

peterdesmet: Try 2015_HOMARUS

PietrH:
@peterdesmet ready for review

- 1 test failing: https://github.com/inbo/etn/blob/78712927c2a13177a24ba58f3f0852818e3800d6/tests/testthat/test-get_acoustic_detections.R#L19

Sometimes it passes, sometimes it does not. Seems to fail when doing `devtools::check()` but pass in console or when running tests for a single file. The test itself seems fine, I suspect there might be duplicate detection_id's in the acoustic detentions table. 

The object that we test here `df` could be different every run, because it uses a limit without an order by statement. 

- Feel free to fix style issues directly, or just let me know.

peterdesmet (approved):
Reviewed, sorry it took so long! 😅 

- I would move `R/testthat-helpers.R` to `tests/testthat/helper.R` (cf. camtraptor)
- Made a number of stylistic changes, but would be good if all are tested again
- Are changes made to the citation now?

PietrH:
Tests pass

PR #294, opened 2024-06-21
Hotfix `write_dwc` failing sql parsing + support uppercase

### LINKED ISSUE #289: `write_dwc()` returns no data without warning or error
When running `write_dwc()` the resulting file contains no data (only the headers):

``` r
library(etn)
con <- connect_to_etn()
data <- write_dwc(con, "2011_RIVIERPRIK")
#> Reading data and transforming to Darwin Core.
#> Writing data to:
#> ./dwc_occurrence.csv
data
#>  [1] type                          license                      
#>  [3] rightsHolder                  datasetID                    
#>  [5] institutionCode               collectionCode               
#>  [7] datasetName                   basisOfRecord                
#>  [9] dataGeneralizations           occurrenceID                 
#> [11] sex                           lifeStage                    
#> [13] occurrenceStatus              organismID                   
#> [15] organismName                  eventID                      
#> [17] parentEventID                 eventDate                    
#> [19] samplingProtocol              eventRemarks                 
#> [21] locationID                    locality                     
#> [23] decimalLatitude               decimalLongitude             
#> [25] geodeticDatum                 coordinateUncertaintyInMeters
#> [27] scientificNameID              scientificName               
#> [29] kingdom                      
#> <0 rows> (or 0-length row.names)
```

<sup>Created on 2023-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PietrH:
- Set extra parameter to ensure passing `NULL` to glue is passed as `"NULL"` strings to the database, instead of the whole query being dropped. 

- Updated a snapshot, because the database has changed since the last unit test check. 

- Convert `animal_project_code` to lowercase for SQL injection in `write_dwc()`: fixes #289 : bff9852f

---

Since glue v1.7.0 a literal NULL will result in the line (in this case, the whole file) being dropped. Before it was returned as "NULL" by default. This change set's it to it's old default behaviour.

In principle this bug would have been present as soon as someone updated glue as of the 10th of January this year.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # etn 2.2.1
 
 * Fix bug in `write_dwc()` where the function would return an error due to an updated dependency (#293).
-- Added support for UPPERCASE `animal_project_code`'s in `write_dwc()` #289
+* Add support for UPPERCASE `animal_project_code`s in `write_dwc()` (#289).
 
 # etn 2.2.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
+# etn 2.2.1
+
 # etn 2.2.0
 
 * Add `NEWS.md` file to communicate changes to the package.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # etn 2.2.1
 
+- Fix bug in `write_dwc()` bug where the function would return an error due to an updated dependency. #293
+- Added support for UPPERCASE `animal_project_code`'s in `write_dwc()` #289
+
 # etn 2.2.0
 
 * Add `NEWS.md` file to communicate changes to the package.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # etn 2.2.1
 
-- Fix bug in `write_dwc()` bug where the function would return an error due to an updated dependency. #293
+* Fix bug in `write_dwc()` where the function would return an error due to an updated dependency (#293).
 - Added support for UPPERCASE `animal_project_code`'s in `write_dwc()` #289
 
 # etn 2.2.0
--- a/R/write_dwc.R
+++ b/R/write_dwc.R
@@ -56,6 +56,8 @@ write_dwc <- function(connection = con,
     length(animal_project_code) == 1,
     msg = "`animal_project_code` must be a single value."
   )
+  ## Set animal project code to lowercase for sql
+  animal_project_code <- stringr::str_to_lower(animal_project_code)
 
   # Check license
   licenses <- c("CC-BY", "CC0")
--- a/R/write_dwc.R
+++ b/R/write_dwc.R
@@ -84,7 +84,8 @@ write_dwc <- function(connection = con,
   message("Reading data and transforming to Darwin Core.")
   dwc_occurrence_sql <- glue::glue_sql(
     readr::read_file(system.file("sql/dwc_occurrence.sql", package = "etn")),
-    .con = connection
+    .con = connection,
+    .null = "NULL"
   )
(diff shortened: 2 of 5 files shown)


PietrH:
@peterdesmet, I also added a fix for #289 , Sorry, I didn't notice you already reviewed. 

If you give the go ahead, I'll merge.

NEWS.md
```diff
@@ -1,3 +1,8 @@
+# etn 2.2.1
+
+- Fix bug in `write_dwc()` bug where the function would return an error due to an updated dependency. #293
```
peterdesmet, line 3:
```suggestion
* Fix bug in `write_dwc()` where the function would return an error due to an updated dependency (#293).
```

Typo ("bug" twice) + issue number is placed between parenthesis before end period. Since `*` is used for list bullets before, I would keep doing that here.

NEWS.md
```diff
@@ -1,3 +1,8 @@
...
+- Added support for UPPERCASE `animal_project_code`'s in `write_dwc()` #289
```
peterdesmet, line 4:
```suggestion
* Add support for UPPERCASE `animal_project_code`s in `write_dwc()` (#289).
```

R/write_dwc.R
```diff
@@ -56,6 +56,8 @@ write_dwc <- function(connection = con,
     length(animal_project_code) == 1,
     msg = "`animal_project_code` must be a single value."
   )
+  ## Set animal project code to lowercase for sql
+  animal_project_code <- stringr::str_to_lower(animal_project_code)
```
peterdesmet, line 60: Are all project codes lowercase in the database then?

tests/testthat/test-write_dwc.R
```diff
@@ -63,3 +63,13 @@ test_that("write_dwc() returns the expected Darwin Core terms as columns", {
     )
   )
 })
+
+test_that("write_dwc() supports uppercase animal_project_code's", {
```
peterdesmet, line 67:
```suggestion
test_that("write_dwc() supports uppercase animal_project_codes", {
```

peterdesmet (changes requested):

R/write_dwc.R
PietrH: I'm relying on the validity of this line in `get_animal_projects()`:

https://github.com/inbo/etn/blob/326d9e8293c51080b87a3b8569f6a3df7291225f/R/get_animal_projects.R#L39-L42

If it's ok to `LOWER()` all codes there, surely it's ok to do the same here?

PietrH: Oh, I see, they are lowered on the database side, and then compared to the client entry. It's not entirely the same. 

We might need a different solution then! How do we handle this for the other functions? We could change the sql for `write_dwc()` I suppose.

peterdesmet: I think it's fine. In the dwc_sql we compare with the lowered code too:

https://github.com/inbo/etn/blob/326d9e8293c51080b87a3b8569f6a3df7291225f/inst/sql/dwc_occurrence.sql#L16

https://github.com/inbo/etn/blob/326d9e8293c51080b87a3b8569f6a3df7291225f/inst/sql/dwc_occurrence.sql#L78

https://github.com/inbo/etn/blob/326d9e8293c51080b87a3b8569f6a3df7291225f/inst/sql/dwc_occurrence.sql#L95

PietrH: In that case I'll go ahead and merge. We can revisit it if someone finds an exception!

PietrH:
Can't merge without an accepted review! Are we ok to merge @peterdesmet ?

peterdesmet (approved):

PR #311, opened 2024-09-24
Update funder, remove logo + custom authors

peterdesmet:


PietrH:
If I run `pkgdown::build_site()` to have a look at the results, I modify a bunch of files. Is it safe/a good idea to commit these changes?

PietrH:
The navbar looks very dark when I build on the VLIZ RStudio server: 

image

PietrH (changes requested):
Could you have a look at the navbar? It seems dark when I build on the RStudio server (pkgdown 2.0.5)

peterdesmet:
> Is it safe/a good idea to commit these changes?

Yes, we currently have to build and commit the website ourselves (not through GitHub Actions). And the current website has errors in all examples, so good to have it updated.

Longer term, this should be done through GitHub Actions on a `gh-pages` branch.

> Could you have a look at the navbar?

Can you install the latest version of pkgdown?

PietrH:
> Can you install the latest version of pkgdown?

Not via `install.packages` because the latest version doesn't support R 3.6.0, I could install it via remotes but I'd have to look into why they dropped support first

PietrH:
I was able to get it to work with `pkgdown` v2.0.9 installed via remotes;

image

PietrH (approved):

PR #313, opened 2024-09-25
Hard deprecate `connect_to_etn()` and reexport

PietrH:
Previously on this dev branch, `connect_to_etn()` was no longer exported, and used internally as a helper. In this PR, a new helper is created to take over this work, and `connect_to_etn()` is re-exported, but hard deprecated: it always returns a lifecycle warning message, and it only returns NULL. All SQL helpers were adapted to use this new helper. Tests were changed over so both `connect_to_etn()` and the new helper are covered.

We decided to do it this way so older scripts that have `con <- connect_to_etn()` at the top are met with a deprecation message rather than a missing function. 

`connect_to_etn()` is no longer neccesairy. Functions that need credentials will prompt the user for them. 

I also added to NEWS and made a small change to a vignette.


PietrH:
Ready for review

R/connect_to_etn.R
```diff
@@ -1,30 +1,19 @@
...
-#' }
-connect_to_etn <- function(username = Sys.getenv("userid"),
-                           password = Sys.getenv("pwd")) {
-  con <- DBI::dbConnect(
-    odbc::odbc(),
-    "ETN",
-    uid = paste("", tolower(username), "", sep = ""),
-    pwd = paste("", password, "", sep = "")
+#' @export
+connect_to_etn <- function() {
```
peterdesmet, line 11: I notice the params have been removed as well. How will this react when a user calls:

```R
connect_to_etn(username = "foo", password = "bar")
# or
connect_to_etn("foo", "bar")
```

peterdesmet (approved):
One comment. Would extend test with function call with parameters.

R/connect_to_etn.R
PietrH: `Error in connect_to_etn("foo", "bar") : unused arguments ("foo", "bar")`

We can get around this with ellipsis, well spotted.

PietrH:
Made requested changes, ready for a second look if you have time. 

If not, let me know and we can merge.

peterdesmet (approved):
I updated the vignette (please verify it's correct) and simplified the test to just make use of `lifecycle::expect_deprecated()`

PietrH:
I'm not a 100% sure how often you'll be prompted for them, but at least every time you click "Quit Session" so I think that's close enough. 

Thanks for the review Peter!

PietrH:
The 2 tests in `connect_to_etn()` still fail:

```
Failure (test-connect_to_etn.R:5): connect_to_etn() returns deprecation warning when used
`connect_to_etn(username = "my name", password = "my password")` did not throw the expected warning.

Failure (test-connect_to_etn.R:8): connect_to_etn() returns deprecation warning when used
`connect_to_etn("my name", "my password")` did not throw the expected warning.
```

I'm looking into it, but would be grateful for any ideas!

R/connect_to_etn.R
```diff
@@ -1,30 +1,21 @@
...
-connect_to_etn <- function(username = Sys.getenv("userid"),
-                           password = Sys.getenv("pwd")) {
-  con <- DBI::dbConnect(
-    odbc::odbc(),
-    "ETN",
-    uid = paste("", tolower(username), "", sep = ""),
-    pwd = paste("", password, "", sep = "")
+#' @export
+connect_to_etn <- function(...) {
+  lifecycle::deprecate_warn(
```
maelle, line 14: unrelated to your question elsewhere: if it is hard deprecated why not use `lifecycle::deprecate_stop()`? it also generates an helpful message.

maelle: (I guess it's because you don't want an error but won't users ignore the warning :sweat_smile: )

R/connect_to_etn.R
```diff
@@ -1,30 +1,21 @@
 #' Connect to the ETN database
 #'
-#' `r lifecycle::badge("deprecated")`
-#'     Connect to the ETN database using username and password.
+#' This function is `r lifecycle::badge("deprecated")` since etn version 2.3.0.
+#' It's use is no longer supported or needed. All connections to the ETN database
```
maelle, line 4:
```suggestion
#' Its use is no longer supported or needed. All connections to the ETN database
```

PietrH:
I documented the behaviour of the tests and added skips for them, I'm now merging so this doesn't block the beta release. 

I'm also planning on creating a minimal test package to see if I can replicate this behaviour, and if so, I'll go looking for help on https://forum.posit.co/

PietrH:
Thank you both for the constructive comments and contributions!

PR #314, opened 2024-09-26
Update examples to use API

PietrH:
- remove `connect_to_etn()` and usage of `connection` argument from all function examples, this function is hard deprecated as of #313 
- Also drop `purrr` as a dependency, is unused. I only recently added it, and apparently refactored out the code where I used it. 

Note that R CMD CHECK will still fail as long as the changes from #313 are not merged and updated into this branch. You can check the package without vignettes by running: `devtools::check(vignette = FALSE)`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,6 @@ Imports:
     lubridate,
     methods,
     odbc,
-    purrr,
     readr,
     rlang,
     stringr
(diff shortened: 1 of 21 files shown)


sannegovaert (approved):

PietrH:
Thank you Sanne!

PR #328, opened 2024-10-17
Use Apache Feather for API object transfer instead of RDS

### LINKED ISSUE #327: API: change compressed output format for get_val to reduce serialisation cost
Currently the `get_val()` helper supports fetching from OpenCPU as JSON or RDS. 

In #323 , Stijn found that we are crashing the session by running out of memory, possibly on a serialisation process. I believe `base::writeRDS()` -> `base::serialize()` might be the cause of this memory usage. Assuming the crash happens on the writing of the object as RDS to the output stream. I've not been able to replicate this issue locally or on the RStudio Server. 

There are a few open issues on opencpu for child processes that died:
- https://github.com/opencpu/opencpu/issues/391
- https://github.com/opencpu/opencpu/issues/383
- https://github.com/opencpu/opencpu/issues/366
- https://github.com/opencpu/opencpu/issues/319
- https://github.com/opencpu/opencpu/issues/245

I did a quick local test on a deployments table to see if outputting as feather or parquet might help:

```
Unit: milliseconds
           expr       min        lq      mean    median
 rw_feather(df)  13.28443  14.40181  15.38457  15.07805
 rw_parquet(df)  38.05100  39.58587  41.90330  40.76735
     rw_rds(df) 263.91068 265.43494 269.96770 266.53872
        uq       max neval
  16.14836  22.61397   100
  42.62479  60.33460   100
 270.30619 296.12905   100
```

It looks like both are faster on my system, I have not benchmarked memory usage yet. 

This is using lz4 compression for feather, snappy for parquet and gzip for rds. 

Stijn proposes using an alternative fetch method; returning the session id and writing out paged result objects to the session dir, then having the client fetch these objects and reserialazing on the client

PietrH:
This PR requires changes on the Lifewatch RStudio Server in order to work on there, I want users on the RStudio Server to be able to use the most recent etn version. **If this PR is merged into v2.3.1, then this version will require an update of the R version on the lifewatch server.** 

This PR is serving as a jumping off point for testing by my co-developers. I'd be very grateful for your input! 

Fixes:
- I now use apache arrow for file transfers instead of RDS, which uses lz4 instead of gzip compression and is also chunked:
    - less memory usage for both client and server
    - faster compression/decompression
    - because we need less memory, we can get away with larger objects

Talking points:
- Doesn't actually allow you to load detections for `2013_albertkanaal` this crashes before serialisation on the server side, so I can't fix this on the client side
- Is it actually faster for you?
- You can test memory usage by looking at your resource manager, or via something like `bench::mark()`
- Is the arrow dependency worth it? RDS will become more and more problematic with larger and larger datasets (especially multiple animal_project_codes)

## Alternative approach
The API result object is currently passed as a single binary stream. Instead, I could also try to split it up into multiple files hosted by OpenCPU and to fetch those individually, and combine them. This is more complex and will be more difficult to debug if something goes wrong, but would allow us to keep using rds for now. I haven't tested this yet, as it would disrupt the current v2.3.0 beta. There is a dev env on the horizon that would allow tests like this in the future.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: etn
 Title: Access Data from the European Tracking Network
-Version: 2.3.0
+Version: 2.3.1.9000
 Authors@R: c(
     person("Peter", "Desmet", email = "peter.desmet@inbo.be",
            role = c("aut", "cre"), comment =  c(ORCID = "0000-0002-8442-8025")),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ License: MIT + file LICENSE
 URL: https://github.com/inbo/etn, https://inbo.github.io/etn
 BugReports: https://github.com/inbo/etn/issues
 Depends:
-    R (>= 3.4.0)
+    R (>= 4.0)
 Imports:
     arrow,
     askpass,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,6 +33,7 @@ Imports:
     lubridate,
     methods,
     odbc,
+    protolite,
     readr,
     rlang,
     stringr
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,6 +21,7 @@ BugReports: https://github.com/inbo/etn/issues
 Depends:
     R (>= 3.4.0)
 Imports:
+    arrow,
     askpass,
     assertthat,
     DBI,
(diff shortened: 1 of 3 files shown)


PietrH:
#329 serves as an anchor for a version upgrade of the R version on the Lifewatch RStudio Server. This will open up many more recent R package versions to the users on the server. This will become more and more relevant as tidyverse drops support for R versions lower as 4.

PietrH:
Blocked by #329

PietrH:
Current idea is to detect if the client supports Apache Feather for object transfer, and if it does, use it. 

Could it support arrow -> Suggest to install it, use RDS
Arrow installed -> use Apache Feather

PR #352, opened 2025-04-17
Add (zenodo) DOI to README

PietrH:
All past releases are now archived to Zenodo #349, every release has it's own DOI. There is also a DOI that refers to all releases, automatically forwarding to the most recent release: https://doi.org/10.5281/zenodo.15235747

This PR adds this DOI to the README page in the form of a badge (Zenodo conveniently already provides the markdown code to do this) by adding it to `README.Rmd` and then building that file to `README.md`. 

By building the README file, I noticed that the R-CMD-CHECK badge was also missing, this badge was automatically added by building. `devtools::build_readme()`. Running this function also changed the formatting slightly. 

## Doubts

~~I updated the `{cffr}` Github action to use a custom key to add the (all version) Zenodo DOI to the  `CITATION.cff file, it'll be automatically added whenever this action runs (on releases).~~

EDIT: I added a line to the `CITATION` file instead, and left the action as is. If there is ever a CRAN release, that second DOI will be added under `identifier`: https://docs.ropensci.org/cffr/news/index.html#cffr-110

However:
- ETN is not on CRAN, so it doesn't have a CRAN DOI.
- If it is ever released to CRAN (hopefully!), it'll get a DOI there, and `{cffr}` is supposed to read it from the `CITATION` file. 
- I'm unsure if the CRAN DOI will be overwritten by the Zenodo DOI.
- The DOI I provided always directs to the most recent version, not neccesairily the version the user is citing when running `citation("etn")`, this might be confusing.

There might be a workaround to get `{cffr}` to update `CITATION.cff` with the Zenodo DOI for the specific release instead of the DOI for all releases, but I didn't investigate this option.

### Changes (AI Summary)

This pull request updates the citation information for the `etn` R package, including the addition of Digital Object Identifiers (DOIs) for the package and its dependencies. It also enhances the documentation to reflect these updates.

* Added a DOI (`zenodo.15235747`) for the `etn` package in the `CITATION.cff` file and the `inst/CITATION` file. [[1]](diffhunk://#diff-9b9f76394a441ff40bf8d7a0f3e0ddd7ae97abfa9a9a6abaedccf76dc4d51295R1-R12) [[2]](diffhunk://#diff-f36d9cd7ba8ef13c4e62d697ddf4a1bd0f0eabb06a36e8bfb37898aee66976dcL10-R11)
[...]

PR #355, opened 2025-04-22
Point badge image to versionless DOI

peterdesmet:
The Zenodo badge added in #352 points to the correct (versionless) DOI `10.5281/zenodo.15235747`, but the badge itself displays the versioned DOI `10.5281/zenodo.15235777` using the image URL https://zenodo.org/badge/95901229.svg (I don't know where the image URL is coming from).

This PR points the badge image to https://zenodo.org/badge/DOI/10.5281/zenodo.15235747.svg (the versionless DOI).


PietrH:
That's a strange quirck of the markdown suggested by Zenodo itself for the badge, their suggestion looks like this:

image

The image indeed doesn't match the link, good catch!

PietrH (approved):

PietrH:
Let's merge after the CI has run

PR #357, opened 2025-04-22
Hotfixes: fix `check_value()` error messaging + remove duplicate `receiver_ids` + `write_dwc(rights_holder = NULL)` sql bug

PietrH:
# Hotfixes

> `check_value()` provides an error when a value is missing from a reference vector. However, currently, when the reference vector contains `NA`, and the assertion fails (which works correctly), the resulting message doesn't list the contents of the reference vector but replaces it by `NA`.

- [x] Add to NEWS
- [x] Update version number: patch
- [x] run tests with db access
- [x] build_site : for upping version number
- [ ] create tag + release after merge to main

### Bug Fixes and Improvements:
* Fixed an issue in `check_value()` where error messages failed to format correctly when `NA` values were present in the reference list. The logic for removing `NA` values was corrected.
* Resolved a bug in `list_receiver_ids()` where `NA` values were included in the results. The function now explicitly filters out `NA` values.
* Addressed a problem in `write_dwc()` where the function would fail if `rights_holder` was not provided. A default value of `NA_character_` is now assigned when `rights_holder` is missing.

### Test Enhancements:
* Added new tests for `check_value()` to verify it handles `NA` values in the reference list and produces appropriate error messages.
* Enhanced tests for `list_receiver_ids()` to ensure it returns a unique, non-`NA` character vector and includes known values.
* Introduced tests for `write_dwc()` to validate the behavior of the `rights_holder` parameter, including cases where it is explicitly set or left as default.

### Documentation and Metadata:
* Updated the package version to `2.2.2` in `DESCRIPTION` to reflect the new changes.
* Added a changelog entry in `NEWS.md` summarizing the fixes and improvements in this release.


PietrH:
Found bug : #358

tests/testthat/test-check_value.R
```diff
@@ -42,3 +42,16 @@ test_that("check_value() can ignore case", {
     c("a", "b")
   )
 })
+
+test_that("check_value() can handle NA in reference",{
+  expect_identical(
+    check_value("A", c("A",NA,"C")),
```
peterdesmet, line 48: Lint my complain about the missing spaces in your vector.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: etn
 Title: Access Data from the European Tracking Network
-Version: 2.2.1
+Version: 2.2.2
```
peterdesmet, line 3: If you bump here, don't forget to bump in CITATION.cff before you release. Same with DOI in CITATION.cff, which is tricky, since you don't know it yet. I have no solution for this.

peterdesmet (approved):
Longest feature branch name ever 😉.

DESCRIPTION
PietrH: I used the general, all version DOI in my citation.cff. Was that a mistake?

peterdesmet: Ah, didn't catch that. That works! 👍

PR #363, opened 2025-04-28
Use ROR for copyright holder

peterdesmet:
[pkgdown 2.1.2](https://github.com/r-lib/pkgdown/releases/tag/v2.1.2) now supports ROR Identifiers for organizations. Since this supports better linked data, I think we should adopt it for our packages.

It will display on the pkgdown website as:

Screenshot 2025-04-28 at 15 13 10

If you have an older version of pkgdown, then it will show as:

Screenshot 2025-04-28 at 15 13 23

I've also updated the format of the authors in DESCRIPTION cf. other packages.


PietrH:
The latest version of `pkgdown` depends on R4.0.0, but the RStudio server is still running R3.6.3 from 5 years ago. An update is in the works, but is a while off. Maybe by the end of the year we could expect an update, we'll see. 

```
platform       x86_64-pc-linux-gnu         
arch           x86_64                      
os             linux-gnu                   
system         x86_64, linux-gnu           
status                                     
major          3                           
minor          6.3                         
year           2020                        
month          02                          
day            29                          
svn rev        77875                       
language       R                           
version.string R version 3.6.3 (2020-02-29)
nickname       Holding the Windsock  
```

In the meantime I'm looking for a workaround for this PR so we can deploy the new pkgdown documentation website in the shorterm. 

In the medium term, after v2.3 we could deploy the pkgdown website via Github Actions on a more recent version of R as the examples and vignettes could build via the API.

PietrH:
Building on the VLIZ RStudio Server results in:

image


Maybe I can build just the index locally?

PietrH:
I'm just going to live with the ROR logo being missing for now

PietrH (approved):
Thanks for picking this up so quickly Peter! Doesn't look perfect at the moment but it'll resolve itself as etn goes to v2.3+ and the github action supports pkgdown v2.1.2

:rocket:

PR #366, opened 2025-05-05
Update test for archival type tags in `get_animals()`

### LINKED ISSUE #365: `get_animals()` now includes `tag_type == "archival"`
A test started failing because of a new `tag_type`, the comment in the test indicates that `archival` tag types used to be excluded. At least since 2011. 

I don't believe this database side change should break anything except this test which should be updated. I did come across a different issue #364.

PietrH:
* The most important change in this PR is a change to a test that failed because `archival` type tags used to be missing from `get_animals()`, this PR thus mainly serves as documentation of this change in the database for etn package users.
* Added this change to `NEWS.md` so etn package users are notified.
* Starting from this PR the version of etn on `main` is using `.9000` as the development version number. As discussed with Peter.
* Fixed a bug in which a test that was supposed to clean up temp dataset files failed to do so leaving a folder behind in `tests/testthat` after a test run.


PietrH:
Now that I think about it, I'm quite surprised that the number of detections in `2014_demer` changed. @peterdesmet , we expect this number to remain stable don't we?

peterdesmet:
Indeed, it should be stable. The changing number might be caused by reused tags 🤔.

PietrH:
I've sent an email to Claudia about this. I'll let the snapshot update in this PR. I prefer that over skipping the test or having it fail. 

We could make the snapshot insensitive to changes in the exact numbers by adding a transform function to the snapshot:

```r
expect_snapshot(
    cat(download_acoustic_dataset(con, "2014_demer")),
     transform = ~stringr::str_replace(.x, "(?<=:)\\s+[0-9]+$", "<number_removed>")
)
```
Or even adding rounding instead of removing the numbers entirely, but in this case I think having the test fail is preferable.

peterdesmet (approved):
All good 👍

PR #370, opened 2025-07-22
Credentials vignette, no longer write to sys env, test `get_credentials()` prompting.

### LINKED ISSUE #335: Store credentials in  `.Renviron`, or switch to `keyring`?
2.  When not inputting the userid correctly (happened to me that I was running several lines of code quickly after each other, and the next line was taken as input), is there a way to make the prompt for userid and pwd reappear? Or is it maybe also possible to store userid and pwd somewhere? (.Renviron does not work, which was expected).

_Originally posted by @lottepohl in https://github.com/inbo/etn/issues/318#issuecomment-2447216295_
            

`{rgbif}` stores user credentials in `.Renviron`, however, making changes automatically to this file could cause unexpected results for users. It is making a change outside of the environment of the package, same with `Sys.setenv()` really. I'm thinking about this section in the R Packages book 2nd edition: https://r-pkgs.org/code.html#sec-code-r-landscape

Storing information here is simple enough: `usethis::edit_r_environ()`, but is this a better solution than `Sys.setenv()` on the long term? 

`{keyring}` is used by `{move2}`: https://gitlab.com/bartk/move2/-/blob/main/R/movebank_credentials.R and is something I had in mind on the long run

```r
  keyring::key_set_with_value(
    username = username, password = password,
    service = key_name, keyring = getOption("move2_movebank_keyring")
  )
```

The working of `usethis::edit_r_environ()` isn't too complex internally:

```r
function (scope = c("user", "project")) 
{
    path <- scoped_path_r(scope, ".Renviron", envvar = "R_ENVIRON_USER")
    edit_file(path)
    ui_bullets(c(`_` = "Restart R for changes to take effect."))
    invisible(path)
}
```
With helper:
```r
function (scope = c("user", "project"), ..., envvar = NULL) 
{
    scope <- match.arg(scope)
    if (scope == "user" && !is.null(envvar)) {
        env <- Sys.getenv(envvar, unset = "")
        if (!identical(env, "")) {
            return(user_path_prep(env))
        }
    }
    root <- switch(scope, user = path_home_r(), project = proj_get())
    path(root, ...)
}
```
And cascades into more and more helpers... But it could be done

### LINKED ISSUE #338: Document that you can set database credentials as sys env
Lotte had not spotted this option, thus it needs to be better documented somewhere!

PietrH:
This PR introduces a new vignette that documents how to set credentials for the ETN package via the `.Renviron` file. It also explicitly removes the behaviour of the `get_credentials()` helper to set these variables. 

Old behaviour:
- If no credentials stored: prompt for them
- store them as env vars

New behaviour: 
- If no credentials stored, prompt for them
- If credentials stored: use these instead. 

It also introduces tests for `get_credentials()` to test the prompting itself, and to make sure that the function doesn't write to env variables. This required the usage of [mocking tools](https://testthat.r-lib.org/reference/local_mocked_bindings.html) within testthat, which in their turn required me to wrap a number of functions:

- rlang::is_interactive()
- askpass::askpass()
- base:readline()

Finally it changes the `get_credentials()` logic to be more strict when it comes to detecting no credentials are stored. If no credentials are stored and the session is not running in interactive mode, an error is returned. 

## TODO
- [x] fix R CMD CHECK warning:
```
* checking R code for possible problems ... WARNING
Warning: Found a .Internal call in the following function:
  ‘prompt_user’
with calls to .Internal functions
  ‘readline’
Packages should not call .Internal(): it is not part of the API, for
use only by R itself and subject to change without notice.
```
- [x] update snapshots


PietrH:
❯ checking R code for possible problems ... WARNING
  Found a .Internal call in the following function:
    ‘prompt_user’
  with calls to .Internal functions
    ‘readline’
  
  Packages should not call .Internal(): it is not part of the API, for
  use only by R itself and subject to change without notice.

❯ checking DESCRIPTION meta-information ... NOTE
Error: Error: R CMD check found WARNINGs
Execution halted
  License stub is invalid DCF.

❯ checking dependencies in R code ... NOTE
  Namespaces in Imports field not imported from:
    ‘askpass’ ‘etnservice’
    All declared Imports should be used.

PietrH:
>Namespace in Imports field not imported from: ‘etnservice’
  All declared Imports should be used.
  
  Not fixed in 237f10a6deec983e10029a14e726196153e22c83 but I'll keep it as it still improves documentation of this rather convoluted helper.

PietrH:
I'm failing on everything except MacOS, works on my laptop. Is there something wrong with vcr? 

I'm getting 403's somewhere?

PR #373, opened 2025-07-23
Refactor to httr2 + Improve error on bad credentials

### LINKED ISSUE #324: Check if provided credentials are correct
I just ran the new `etn` beta version on my local drive and as explained by @PietrH I had to provide my credentials after running a function. However, I still get an error when running a function, for example `get_animal_projects()`:

```
> get_animal_projects()
Error: ODBC failed with error 00000 from \
             .

In call:
opencpu:::rapachehandler()

Backtrace:
    ▆
 1. └─output_handler$error(e)
```

When filling in my password, I did not include the `"` showing its a factor. Could that be the issue or would the password prompt appear again when filling in the incorrect password?

### LINKED ISSUE #376: Update NEWS
Add the changes from #370 to NEWS.md in v2.3.1

> This PR introduces a new vignette that documents how to set credentials for the ETN package via the `.Renviron` file. It also explicitly removes the behaviour of the `get_credentials()` helper to set these variables.
> 
> Old behaviour:
> 
>     * If no credentials stored: prompt for them
> 
>     * store them as env vars
> 
> 
> New behaviour:
> 
>     * If no credentials stored, prompt for them
> 
>     * If credentials stored: use these instead.
> 
> 
> It also introduces tests for `get_credentials()` to test the prompting itself, and to make sure that the function doesn't write to env variables. This required the usage of [mocking tools](https://testthat.r-lib.org/reference/local_mocked_bindings.html) within testthat, which in their turn required me to wrap a number of functions:
> 
>     * rlang::is_interactive()
> 
>     * askpass::askpass()
> 
>     * base:readline()
> 
> 
> Finally it changes the `get_credentials()` logic to be more strict when it comes to detecting no credentials are stored. If no credentials are stored and the session is not running in interactive mode, an error is returned.
> ## TODO
> 
>     * [x]  fix R CMD CHECK warning:
> 
> 
> ```
> * checking R code for possible problems ... WARNING
> Warning: Found a .Internal call in the following function:
>   ‘prompt_user’
> with calls to .Internal functions
>   ‘readline’
> Packages should not call .Internal(): it is not part of the API, for
> use only by R itself and subject to change without notice.
> ```
> 
>     * [x]  update snapshots

### LINKED ISSUE #378: Refactor API code to use httr2 instead
Used to use httr, now use httr2

PietrH:
- Should I change etnservice so `validate_login` accepts a `credentials` object and thus get rid of the need to use a custom request? inbo/etnservice#79
- [x] Consider improving error message:
```r
glue::glue("Failed to login with username {get_credentials()$username}",
           ", please try again. If you have stored your credentials, change them in .Renviron via usethis::edit_r_environ()")
```

cli would make this much better.

Currently introducing an undocumented (but really already indirect) dependency on httr2

I wonder how hard it would be to drop the httr dependency and refactor for httr2

- [x] update news
- [x] forward R errors to client
- [x] change `forward_to_api()` to optionally not append credentials? 
Then I wouldn't need the custom request for validate credentials anymore.
- ~~change `etnservice::validate_credentials()` to accept credentials object in `...`
That way we can just use the forward to api flow instead of a custom request just for this function.~~
- [x] Fix so user is only prompted for credentials once per function call


PietrH:
If you have no credentials stored, are you prompted twice per function call? (once for the function, once to validate?)

Issue #374, opened 2025-07-24
Switch to remotes: `ropensci/vcr` for vcr v2.0.0


PietrH:
As mentioned in ropensci/vcr#539 the CI failures in #371 might be fixed in vcr v2.0.0

Because this version of vcr is not out on CRAN yet, switching to ropensci (github) as a remote might fix the issue. 

I believe it's probably a good idea to set a minimum version for vcr as well.

PR #375, opened 2025-07-24
Switch to vcr v2.0.0, not out on CRAN yet

PietrH:
See https://github.com/inbo/etn/issues/374#issue-3258876539

Switching to remotes to get a very recent version of vcr since versions lower as v2 break all cassettes when new versions of httr2 are installed on the testing system. 

When vcr v2.0.0 is released on CRAN, we can switch away from `Remotes: ropensci/vcr`

- [x] Turn VCR back on in Repo actions variable
- [x] Rerun R CMD CHECK with VCR turned on 
- [ ] Document/Create issue to switch back to CRAN version of vcr when v2.0.0 is out
- [x] Consider refreshing all cassettes. 
- [ ] Update news


PietrH:
There is limited impact on users, as they normally don't install Suggests, and devs will probably already be running quite updated systems. 

I see this as a better solution as limiting the package to a specific older release of httr2, especially since there are plans of eventually moving away from httr to httr2.

PR #381, opened 2025-07-29
Add R CMD CHECK runners: old R versions.

### LINKED ISSUE #380: Test on old R version: R3.6.3
The VLIZ RStudio Server currently runs R3.6.3, I should probably test for this version, and the minimum supported version of the R package. 

It might be wise to also test for the OS that the VLIZ RStudio is running on. 

This is due to change with a future update to the VLIZ RStudio version to a more recent R version.

PietrH:
`{etn}` is used on some systems running old R versions, and the package declares a minimum version of R3.4.0, the VLIZ Rstudio Server is on R3.6.3.

This PR adds R CMD CHECK runners for those versions, so we have some security going forward that it actually works for these old R releases. 

I couldn't set the OS to the same one used on the RStudio Server, because such old runners are no longer supported.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -24,6 +24,8 @@ jobs:
           - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
           - {os: ubuntu-latest,   r: 'release'}
           - {os: ubuntu-latest,   r: 'oldrel-1'}
+          - {os: ubuntu-latest,   r: '3.6.3'}
+          - {os: ubuntu-22.04,   r: '3.4.0'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


PietrH:
This rules out vcr v2.0.0, which is supposed to fix the bug with local httr2 installs. 

Causing a bit of a circular argument, I can't require vcr 2.0.0 because then I drop support for R3.6.3 on the VLIZ RStudio Server, but I can't lower my requirement in case users have a recent version of httr2 installed...

This PR is blocked for now.

PietrH:
It might be a solution to drop support for R3.6.3 as the VLIZ RStudio Server is supposed to update to R4.1.0 or higher.

.github/workflows/R-CMD-check.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
           - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
           - {os: ubuntu-latest,   r: 'release'}
           - {os: ubuntu-latest,   r: 'oldrel-1'}
+          - {os: ubuntu-latest,   r: '3.6.3'}
```
peterdesmet, line 28: I suggest adding a comment above this line, cf. https://github.com/inbo/movepub/blob/1d4d7d0d6cef62203c02c4a5665ce538e8a8a4c1/.github/workflows/R-CMD-check.yaml#L28

```
# And minimum supported version in DESCRIPTION
```

PietrH: Good idea, I've added it!

PR #383, opened 2025-08-20
Client can switch to etnservice test deployment.

### LINKED ISSUE #362: Allow ETN API client to switch between test and production domain, without exposing test domain
Currently the production domain is hardcoded into the ETN api client. This is very convenient for everyone except developers (me) trying to test the test branch of etnservice. 

After some discussion with other devs I think the best option is to add a check in `forward_to_api()`

https://github.com/inbo/etn/blob/7f9e23aa54b0b638bf21c0268d5ae12556ab8c49/R/utils.R#L285C1-L285C15

If I add a `Sys.getenv("USE_TEST_API") != ""` check, so that the production domain can be overwritten with a sys env, we can even have a seperate CI run for the test api, all without exposing the url of the test domain (as requested by VLIZ, understandably).

PietrH:
`etnservice` can be automatically deployed to a test instance of the API hosted at a secret url. To develop new functions in `etn` and to test the functionality of `etnservice` it is convenient to quickly and easily switch between the production and test deployments. 

This PR adds a `Sys.getenv()` call to `forward_to_api()` so that if this environmental variable is set, the value of this variable is used instead of the production API endpoint: https://opencpu.lifewatch.be/library/etnservice/

This PR also makes a small change to `forward_to_api()` so you can just omit the `payload` argument if you intend to send a request without a payload. 

This PR is merged into the v2.3.1 branch, which is the branch I'm developing the next major release of ETN on. It does not go into production just yet.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 * Queries via the API and on the Lifewatch RStudio Server will now always return the same results. (#317)
 * You can now store your password and username in `.Renviron` (easy to edit with `usethis::edit_r_environ()`), specifically in `ETN_USER` and `ETN_PWD` (#317, #339, #338, #228)
 * Archival tags are now available in `get_animals()` (#365). 
+* Contributors can now change the default domain of the API to the url of a test deployment by setting the environmental variable `ETN_TEST_API`. (#383)
 
 # etn 2.2.2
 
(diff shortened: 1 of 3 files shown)


PietrH:
HTTP 502 is an OpenCPU code for Nginx (opencpu-cache) can't connect to OpenCPU server, did we overwhelm the API with this CI run?

PR #384, opened 2025-08-20
`get_acoustic_detections()` can now fetch more detections, with less duplicates.

### LINKED ISSUE #323: Downloading data: child process has died
I got the following error when trying to download the largest dataset I know:

```R
> download_acoustic_dataset(animal_project_code = "2013_albertkanaal")
Downloading data to directory `2013_albertkanaal`:
* (1/6): downloading animals.csv
* (2/6): downloading tags.csv                                                                   
* (3/6): downloading detections.csv                                                             
Error: child process has died

In call:
tryCatch({
    if (length(priority)) 
        setpriority(priority)
    if (length(rlimits)) 
        set_rlimits(rlimits)
    if (length(gid)) 
        setgid(gid)
    if (length(uid)) 
        setuid(uid)
    if (length(profile)) 
        aa_change_profile(profile)
    if (length(device)) 
        options(device = device)
    graphics.off()
    options(menu.graphics = FALSE)
    serialize(withVisible(eval(orig_expr, parent.frame())), NULL)
}, error = function(e) {
    old_class <- attr(e, "class")
    structure(e, class = c(old_class, "eval_fork_error"))
}, finally = substitute(graphics.off()))
```

This type of time-outs is expected when using the API. Is there an option to catch they and suggestion something more helpful?

### LINKED ISSUE #340: include `deployment_id` as arg for `etn::get_acoustic_detections()`
Hi everyone,

at the moment, acoustic detections can be queried using the following arguments:

get_acoustic_detections(
  connection = con,
  start_date = NULL,
  end_date = NULL,
  acoustic_tag_id = NULL,
  animal_project_code = NULL,
  scientific_name = NULL,
  acoustic_project_code = NULL,
  receiver_id = NULL,
  station_name = NULL,
  limit = FALSE
)

It would be practical to include the field `deployment_id` in this list. This is relevant to check detections for specific deployments.

All the best,
Lotte

### LINKED ISSUE #382: Switch `get_acoustic_detections()` to client side pagination
## Client side operations:
- sorting the combined result
- The client places one request for a count, this is used for a progress bar and to set the page size
- The client places one request per page, client side pagination
- The page size needs to switch: default 100.000 except when there are less than 100.000 records returned: then ALL (faster). If the result object is very big, then set page size to 1.000.000, say over 2M?
- no longer return `depth_in_meters`: add to NEWS
- new argument: `deployment_id`: #340 
- inbo/etnservice#87
- test for duplicates, last element of previous page might be first element of current page due to implementation of `next_id_pk`

PietrH:
`get_acoustic_detections()` now uses 2 new views and switches between them (via `etnservice::get_acoustic_detections_page()`), and fetches results in a paginated manner. 

This PR also adds a new variable: `detection_id` to filter on. 

`get_acoustic_detections()` loses around 1s on 10 pages with 100.000 records, by waiting on `validate_login()`. It might be worth memoisating this, or refactoring so it's not checked every request. -> I think the time lost on `validate_login()` is actually much greater, up to a few seconds per API call. 

```r

microbenchmark::microbenchmark(validate_login(), times = 50)
Unit: milliseconds
             expr      min       lq    mean   median       uq      max
 validate_login() 960.9077 1127.439 1279.57 1228.272 1330.559 2994.577
 neval
    50
```

## TODO

- [x] complete roxygen2 docs: detection_id
- [x] implement limit
- [x] implement auto page size logic
- [x] refactor; make repeat loop dry; place api logic inside repeat loop?
- [x] client side sorting
- [x] fix error forwarding
- [x] Consider refactoring `validate_login()` -> call much less often.
- [x] Make progress reporting optional
- [x] Capture HTTP 502 and HTTP 503: server side error, ask to try again or report
- [x] fix failing tests: duplicate detections : https://github.com/inbo/etnservice/issues/95 -> missing fields, emailed Stijn
- [x] fix failing tests: download_acoustic_dataset()
- [x] make `forward_to_api()` less repetitive: api response error handling is repeated
- [ ] consider dropping `validate_login()` as default behavior (and dropping memoise and cachem) in favour of better error message handling on etnservice on a bad login.
- [x] CI RUN: test deployment api
- [x] remove the test api var from CI env
- [x] CI RUN: production deployment api : blocked by the deployment of etnservice v0.4.1 to production API


R/get_acoustic_detections.R
```diff
@@ -70,20 +71,140 @@
...
+  # Control number of objects to fetch per page, 100k default, up to 1M for
+  # big queries
+  page_size <- dplyr::case_when(
+    limit ~ 100,
+    n_records_returned > 5e6 ~ 1e6,
+    .default = 100000
+  )
+
+  # Init object to store pages
+  combined_results <- list()
```
PietrH, line 168: It's much more efficient to create a list of known length and then subscript to list elements instead of dynamically growing the list. 

Then again, the length of the list is limited so the actual performance hit on this is probably relatively minor.

PietrH:
While the pagination of a single query has to be done sequentially, I can't request the next page before the previous one has been fetched. Queries over multiple projects can be done in parallel, each with a sequential job per project. 

Might be worth it for a future performance increase.

PietrH:
Failing test due to missing expected receiver id:

```r

Error in `get_acoustic_deployments()`: Can't find receiver_id `POD-3610` in: ...

PietrH:
This test currently fails on main, I think the receiver got removed. Best to find a different CPOD example just to make sure.

PietrH:
`{vcr}` requires R4.1, upping minimum required R version. This was in the works anyway.

PR #389, opened 2025-09-10
Update vcr cassettes, turn response caching back on

### LINKED ISSUE #371: CI: vcr casettes fail to read under httr2 1.2.0
## Bug in `get_val()`: readRDS(.) : unknown input format -> breaks API requests on CI?
- I've got a number of CI runs failing, I believe the problem has started somewhere between my last runs 2 months ago and now. I don't think I introduced it during my development as it started failing on the first runs. 

I've tried refreshing my vcr cache, and turning vcr off. 

- I'm also getting HTTP 403 Forbidden, this might be unrelated. 
- `extract_temp_key()` is also failing on CI, which doesn't use `get_val()` does it? 
- It's somehow passing on MacOS

I believe the response object fed to extract_temp_key() and get_val() is not formed correctly, it contains something unexpected. 

#### get_val()
```r
api_url <- "https://cloud.opencpu.org/ocpu/library/stats/R/rnorm"
httr::POST(api_url, body = list(n = 10, mean = 5)) %>%
  extract_temp_key()
httr2::request("https://cloud.opencpu.org/ocpu") %>%
  httr2::req_url_path_append("tmp", "x00be0faa69f207", "R", ".val","rds") %>%
  httr2::req_perform() %>%
  httr2::resp_body_raw() %>%
  rawConnection() %>%
  gzcon() %>%
  readRDS()
```

I can force a similar error message by requesting the print instead of the rds output format, but trying to parse as rds.

What's confusing me is that it seems to be failing on stored cassettes as well.

PietrH:
Something has been fixed at vcr, or the problem has fixed itself. Anyway, by refreshing the cassettes the unit tests can run again using HTTP response caching greating speeding things up and reducing load on the API when testing. 

Other considerations; 

- [ ] Consider using `transform` to censor the api url from requests, the tmp key is not informative and there is a risk for leaking the test api URI
- [x] Add a few more cassettes for uncovered requests, some are still quite slow. I'm thinking mostly about detection requests. 
- [ ] Make sure all tests pass when using a local sql connection (on R4)
- [x] Consider using `vcr::local_cassette()` as is now recommended: https://docs.ropensci.org/vcr/reference/use_cassette.html
- [x] Consider switching to `qs2` serializer, it's faster and more space efficient. But less transparent in PRs. However, the cassettes are very big and contain binary content anyway for most requests, so this isn't a very big trade-off. 
- [x] skip `validate_login()` on testing?
- [ ] `get_acoustic_detections()` is failing on local queries, it seems to fail on sending `credentials` to `etnservice::get_acoustic_detections_page()`
- [ ] Double check that functions other than `get_acoustic_detections()` can place local sql queries

## Bug

I noticed that `get_acoustic_detections()` doesn't actually work for localdb queries, because the arguments are all passed as a single list into the first (credentials) argument, instead of expanded list. 

The least complicated solution is to switch to `conduct_parent_to_helpers()` instead of a direct call to `forward_to_api()`, but in this case `etnservice::get_acoustic_detections_page()` will need to be renamed to `etnservice::get_acoustic_detections()` and the old function retired. This change might break scripts for users participating in the beta testing until they switch to a newer branch of `{etn}`.

I propose I handle this in a separate issue.


PietrH:
Censoring the url is a bit of a pain, maybe I should just be careful for now. Cassettes are not re-recorded very often anyway.

PietrH:
production deployment of OpenCPU is offline. Reported to Stijn and Paul. Test deployment is working. I'm merging this so I can work on #390 without diverging.

PR #421, opened 2025-10-02
Test that API and SQL queries return identical results

### LINKED ISSUE #330: Test if API and local SQL queries result in same objects
During #317 I noticed 9 functions return a different result between the API and a local SQL query. 

- [x] Add tests all exported functions to see if API and local queries match
- [x] Create custom expectation for comparing api to sql query
- [x] find why there is a mismatch, and fix it!

Some of these differences _might_ be fixed, or at least different after merging #328

## API - SQL mismatch

mismatches for:

- get_cpod_projects
- list_animal_project_codes
- list_acoustic_project_codes
- get_animals
- list_receiver_ids
- get_tags
- get_animal_projects
- list_scientific_names
- get_acoustic_projects

---

              Maybe this is a good use case for a [custom expectation](https://testthat.r-lib.org/articles/custom-expectation.html).

For example, you could create a `expect_call_agnostic()`, with one argument: a function call. The helper could grab the function call (and its arguments), run it locally and over API and expect the result to be the same.

It could then be called as part of a regular test-function.R file:

```R
testthat("get_animals() returns same result locally and via API", {
  expect_call_agnostic(get_animals(animal_id = "5"))
})

_Originally posted by @peterdesmet in https://github.com/inbo/etn/issues/317#issuecomment-2421841943_

### PR DESCRIPTION
**Pull request by PietrH**
# Summary

I've added tests to check for differences between returned objects from functions that run via a local database connection vs via the opencpu API. 

These tests fail. 

Apart from this, the public opencpu instance (also used by ropensci and cran by the way), is currently returning a HTTP 526 error due to a SSL certificate issue. Thus failing two more tests.

# TODO:

I'm splitting off the work to actually fix the differences to a different issue/PRs.

~~Because all these queries are live, both over the api and the database (they have to be, or I'd have to mock all database returns, and then the test is kind of mute), this puts a lot of load on the API.~~

~~Perhaps these tests should only run on a single runner? Or only every so often?~~ 

~~Maybe they should run regardless of changes in the package, on a schedule instead?~~

These tests will only ever run on a machine with local database access, and thus can only be tested on the RStudio server. It will always be skipped on CI. 

- [x] Run the tests on the RStudio Server
Blocked by #423 
- [x] Investigate API - SQL mismatches
- [x] Switch etnservice over to natural sorting via stringr inbo/etnservice#116: implemented in etnservice v0.5.1
- [ ] ~~Add schema for arrow so column classes are consistent for `get_acoustic_detections()`: create via `datapackage.json`? Some booleans are returned as characters over the API (entirely empty) ~~
- [ ] API returns wrong class for moratorium column in several functions:
```
expect_protocol_agnostic(get_cpod_projects())

 Error: api (`actual`) not identical to sql (`expected`).

`actual$moratorium` is a character vector ('0', '1', '1', '1', '0', ...)
`expected$moratorium` is a logical vector (FALSE, TRUE, TRUE, TRUE, FALSE, ...)
 ```

- [ ] Can i just cast all function output based on `datapackage.json` instead of trying to work with R -> Arrow Schema -> R ? 

#### Vector order
Seems to be related to the order in which it is returned. Strange that this differs! Switch `etnservice` to `stringi` natural sorting to avoid OS differences?

Functions that call `stringr::str_sort(data$serial_number, numeric = TRUE)` in `etnservice` do pass. That must be it.

- fixed in inbo/etnservice v0.5.1

#### Boleans as strings

Some fields are interpreted as character, while they are logical. This happens over the API only, it's happening in the OpenCPU layer of the protocol.

```


> # get_acoustic_detections -------------------------------------------------
> 
> # get_acoustic_detections() has internal logic, and should be tested in more
> # detail.
> test_that("get_acoustic_detections() returns identical results on limit", {
+   expect_protocol_agnostic(get_acoustic_detections(limit = TRUE))
+ })
── Failure: get_acoustic_detections() returns identical results on limit ───────────────────────
api (`actual`) not identical to sql (`expected`).

`actual$qc_flag` is a character vector (NA, NA, NA, NA, NA, ...)
`expected$qc_flag` is a logical vector (NA, NA, NA, NA, NA, ...)
Backtrace:
    ▆
 1. └─expect_protocol_agnostic(get_acoustic_detections(limit = TRUE))
 2.   └─testthat::expect_identical(...) at tests/testthat/helper-expectations.R:38:3

Error:
! Test failed
> test_that("get_acoustic_detections() returns identical results on big query", {
+   expect_protocol_agnostic(
+     get_acoustic_detections(animal_project_code = "2011_Warnow")
+   )
+ })
── Failure: get_acoustic_detections() returns identical results on big query ───────────────────
api (`actual`) not identical to sql (`expected`).

`actual$qc_flag` is a character vector (NA, NA, NA, NA, NA, ...)
`expected$qc_flag` is a logical vector (NA, NA, NA, NA, NA, ...)
Backtrace:
    ▆
 1. └─expect_protocol_agnostic(get_acoustic_detections(animal_project_code = "2011_Warnow"))
 2.   └─testthat::expect_identical(...) at tests/testthat/helper-expectations.R:38:3

Error:
! Test failed
> test_that("get_acoustic_detections() returns identical results for multiple stations", {
+   expect_protocol_agnostic(
+     get_acoustic_detections(
+       acoustic_tag_id = "A69-1601-16130",
+       station_name = c("de-9", "de-10")
+     )
+   )
+ })
── Failure: get_acoustic_detections() returns identical results for multiple stations ──────────
api (`actual`) not identical to sql (`expected`).

`actual$qc_flag` is a character vector (NA, NA, NA, NA, NA, ...)
`expected$qc_flag` is a logical vector (NA, NA, NA, NA, NA, ...)
Backtrace:
    ▆
 1. └─expect_protocol_agnostic(...)
 2.   └─testthat::expect_identical(...) at tests/testthat/helper-expectations.R:38:3

Error:
! Test failed
> expect_protocol_agnostic(get_cpod_projects())

 Error: api (`actual`) not identical to sql (`expected`).

`actual$moratorium` is a character vector ('0', '1', '1', '1', '0', ...)
`expected$moratorium` is a logical vector (FALSE, TRUE, TRUE, TRUE, FALSE, ...)
 

> expect_protocol_agnostic(get_animal_projects())

 Error: api (`actual`) not identical to sql (`expected`).

`actual$moratorium` is a character vector ('0', '0', '0', '0', '0', ...)
`expected$moratorium` is a logical vector (FALSE, FALSE, FALSE, FALSE, FALSE, ...)
 

> expect_protocol_agnostic(get_acoustic_projects())

 Error: api (`actual`) not identical to sql (`expected`).

`actual$moratorium` is a character vector ('0', '0', '0', '0', '0', ...)
`expected$moratorium` is a logical vector (FALSE, FALSE, FALSE, FALSE, FALSE, ...)
 

> expect_protocol_agnostic(get_animals())
> expect_protocol_agnostic(get_tags())

 Error: api (`actual`) not identical to sql (`expected`).

`actual$floating` is a character vector (NA, NA, NA, NA, NA, ...)
`expected$floating` is a logical vector (NA, NA, NA, NA, NA, ...)


PietrH:
tests fail on windows, no real output message. This might be present on v2.3.1, something to look into.

PietrH:
`etnservice` seems to return the columns as a boolean, not as `0` or `1`. So this is either an OpenCPU or a client bug. 

~~It's remarkable that it only happens when the serializer is `feather`.~~ Happens on others too.

PietrH:
The `download_acoustic_dataset()` cassette is broken, maybe because of multiple requests.

PietrH:
Switched from `ubuntu-latest` to `macos-latest` for the R 4.1.0 runner, now failing on installing classInt. (installing from source is apparently fragile: r-spatial/classInt#40)

PR #437, opened 2026-01-20
Clean up news

### LINKED ISSUE #412: Clean up NEWS
There are a lot of changes in this milestone. I should clean up NEWS so similar changes are grouped, and it reads a little easier.

- [ ] Check behavior when no credentials are stored, does it only prompt once per session?

PietrH:
I've cleaned up NEWS for the new release so it'll be easier to read.


peterdesmet (changes requested):
Impressive amount of work!

- I've reworked the news order
- I've added installation instructions
- Please tackle https://github.com/inbo/etn/issues/454 so we don't miss anything
- I think this news will go a long way towards a blog post. We might even do a short one like this https://oscibio.inbo.be/blog/movepub-0-4-0/

peterdesmet (approved):
Changing review status to approved, since #454 is best tackled separately.

PR #450, opened 2026-01-29
switch to local downloading feather files

PietrH:
Switch to writing to tempfile then reading: arrow needs random access older versions of arrow do not have this built in.

Setting the min arrow version to 12 would also fix this, but the windows runner for R 4.1.0 can't install arrow 23.0.0 (latest as of writing).


PietrH:
Damiano tested this branch and had no issues on windows 11 desktop.

PR #451, opened 2026-01-29
Add Windows 11 ARM runner to R-CMD-check

PietrH:
This windows arm runner is a full desktop runner, this might resolve any issues that we have trouble reproducing due to differences between windows server and windows desktop.

https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/


PietrH:
The windows-arm runner uses the arm release of R, this release does not have a CRAN binary for arrow. It fails to build arrow from source. I don't think this is worth bugfixing for now. 

We can retry a desktop windows runner later.

PR #452, opened 2026-01-30
Add test for presence of backtrace in API forwarded errors

### LINKED ISSUE #414: VLIZ: turn off default backtrace of `OpenCPU`
Currently R errors are forwarded via OpenCPU with full backtrace, this is not informative for users, especially because actual database errors are omitted. 

I found this in the OpenCPU NEWS file:

https://github.com/opencpu/opencpu/blob/80ea353c14c8601f51ed519744149411d9cc3309/NEWS#L5-L7

```
2.2.11
  - Errors raised by R now by default include a backtrace if available. You can
    opt-out by setting 'error.backtrace' to false in your user.conf.

```

It seems you can turn off this backtrace behaviour by changing a value in the `user.conf` file. This is not documented in the server manual: https://opencpu.github.io/server-manual/opencpu-server.pdf as far as I can see. 

The client package actually nicely forwards R errors, so excluding the backtrace from the error would make the difference between:

**before:**
```
Error in `get_acoustic_detections_page()`:
! Can't find acoustic_tag_id `A69-1602-47652` in: 416kHz-485, 416kHz-956, 416kHz-1141, 416kHz-1164, 416kHz-1277, 416kHz-1935, 416kHz-2066, 416kHz-2219, 416kHz-2228, 416kHz-2248, 416kHz-2485, 416kHz-2681, 416kHz-2718, 416kHz-2961, 416kHz-3305, 416kHz-3572, 416kHz-3696, 416kHz-4017, 416kHz-4155, 416kHz-4475, 416kHz-4731, 416kHz-5045, 416kHz-5079, ...
Backtrace:
     ▆
  1. ├─evaluate::evaluate(...)
  2. │ ├─base::withRestarts(...)
  3. │ │ └─base (local) withRestartList(expr, restarts)
  4. │ │   ├─base (local) withOneRestart(withRestartList(expr, restarts[-nr]), restarts[[nr]])
  5. │ │   │ └─base (local) doWithOneRestart(return(expr), restart)
  6. │ │   └─base (local) withRestartList(expr, restarts[-nr])
  7. │ │     └─base (local) withOneRestart(expr, restarts[[1L]])
  8. │ │       └─base (local) doWithOneRestart(return(expr), restart)
  9. │ ├─evaluate:::with_handlers(...)
 10. │ │ ├─base::eval(call)
 11. │ │ │ └─base::eval(call)
 12. │ │ └─base::withCallingHandlers(...)
 13. │ ├─base::withVisible(eval(expr, envir))
 14. │ └─base::eval(expr, envir)
 15. │   └─base::eval(expr, envir)
 16. │     └─etnservice::get_acoustic_detections_page(...)
 17. │       └─etnservice:::check_value(...)
 18. │         └─assertthat::assert_that(...)
 19. │           └─base::stop(assertError(attr(res, "msg")))
 20. └─evaluate (local) `<fn>`(`<assrtErr>`)
 21.   └─watcher$push(cnd)
 22.     └─handler$error(value)

ℹ This is an error forwarded via the API.
```

after: 

```
Error in `get_acoustic_detections_page()`:
! Can't find acoustic_tag_id `A69-1602-47652` in: 416kHz-485, 416kHz-956, 416kHz-1141, 416kHz-1164, 416kHz-1277, 416kHz-1935, 416kHz-2066, 416kHz-2219, 416kHz-2228, 416kHz-2248, 416kHz-2485, 416kHz-2681, 416kHz-2718, 416kHz-2961, 416kHz-3305, 416kHz-3572, 416kHz-3696, 416kHz-4017, 416kHz-4155, 416kHz-4475, 416kHz-4731, 416kHz-5045, 416kHz-5079, ...

ℹ This is an error forwarded via the API.
```

PietrH:
Added a simple test to see if backtrace information is included in API forwarded errors. 

Tests #414, #477

- [x] Set repo secret to use test api 
- [x] Turn off vcr HTTP response caching
- [x] Turn vcr back on for CI
- [x] change back to production etnservice deployment
- [x] test on production etnservice deployment
- [x] Have CI handle missing value for ETN API URL or remove it from repo secret

This is not a blocker for the ETN API release 

This config change has been pushed to etnservice production, so we should be able to test this on the production API.


PietrH:
The test deployment is still crashing on memory issues: https://github.com/inbo/etn/actions/runs/21942585345/job/63371858144?pr=452

PietrH:
etnservice test is no longer failing on memory issues, but it might start failing again if the load on the API server increases.

PietrH:
I can not locally replicate the test failure in the codecov CI runner. Which is indeed the test that checks for the backtrace:

```r
test_that("opencpu should not pass backtrace", {
  skip_if_offline("opencpu.lifewatch.be")

  # Errors from the API should be forwarded, but backtrace should not be.
  error_message <- expect_error(
    forward_to_api("get_animals",
                   payload = list(animal_id = "not_an_animal_id"),
                  )
  )
  expect_no_match(error_message$message, "Backtrace")
})
```

This is probably because the codecov runner doesn't switch to the test API because it doesn't have access to the repo secret that is needed to get the test API url, thus it falls back to the production API deployment.

PietrH:
> I can not locally replicate the test failure in the codecov CI runner. Which is indeed the test that checks for the backtrace:
> 
> ```r
> test_that("opencpu should not pass backtrace", {
>   skip_if_offline("opencpu.lifewatch.be")
> 
>   # Errors from the API should be forwarded, but backtrace should not be.
>   error_message <- expect_error(
>     forward_to_api("get_animals",
>                    payload = list(animal_id = "not_an_animal_id"),
>                   )
>   )
>   expect_no_match(error_message$message, "Backtrace")
> })
> ```
> 
> This is probably because the codecov runner doesn't switch to the test API because it doesn't have access to the repo secret that is needed to get the test API url, thus it falls back to the production API deployment.

This is indeed the case! And since the change we are testing on is only present on the test deployment, codecov fails. I think this is a desirable behavior so I don't feel the need to change it at the moment.

PietrH:
repo config reset to use vcr and production api as preparation for ETN API release.

PietrH:
@sannegovaert , this is no longer ready for review. I need to investigate test failures first. I'll re-request review when I'm ready again.

PietrH:
The `ETN_TEST_API` repo secret was removed, this might cause the curl parsing error?

PietrH:
> The `ETN_TEST_API` repo secret was removed, this might cause the curl parsing error?

Nope, that's not it!

PietrH:
It seems all API functions are failing to form correct urls, nothing in this PR should be different to main in that aspect. Why is it only failing here? 

```r
get_etnservice_version() lists available functions of etnservice ──
Error in `curl::curl_parse_url(req$url, decode = FALSE)`: Failed to parse URL: Malformed input to a URL function
Backtrace:
     ▆
  1. └─etn (local) `<fn>`()
  2.   └─etn:::forward_to_api(...) at etn/R/utils-api.R:232:3
  3.     ├─httr2::req_retry(...) at etn/R/forward_to_api.R:55:3
  4.     │ └─httr2:::check_request(req) at httr2/R/req-retries.R:101:3
  5.     │   └─httr2:::is_request(req) at httr2/R/req.R:76:5
  6.     ├─httr2::req_body_json(...) at etn/R/forward_to_api.R:55:3
  7.     │ └─httr2:::check_request(req) at httr2/R/req-body.R:103:3
  8.     │   └─httr2:::is_request(req) at httr2/R/req.R:76:5
  9.     └─httr2::req_url_path_append(...) at etn/R/forward_to_api.R:55:3
 10.       └─curl::curl_parse_url(req$url, decode = FALSE) at httr2/R/req-url.R:102:3

```

I need a sanity check, is main broken?

PietrH:
It's got something to do with the test api url injection, when I add the url secret back in, I get a 403 forbidden. Did I paste the wrong url?

PietrH:
The situation is this, it works when a valid url is set to ETN_TEST_API, but fails otherwise. Since main doesn't seem to suffer form this issue (and it has the same CI), I believe the issue lies in the situation where the workflow has access to a system environmental variable, but it's unset in the repo. 

I think removing variable entirely from the workflow would work, but I think it would be even better if the package would cope correctly with this status. The first step is replicating this locally. 

What happens if you set ETN_TEST_API to a random string that is not a valid url?

sannegovaert (approved):
Honestly, I don't understand much of this, but it works so it should be good 😄

PietrH:
Just FYI.

When you pass a function call to the API that results in an error, the etn R package will nicely parse and return this error to you. This test checks if the forwarded error includes a backtrace. 

I don't want to include it, because the backtrace is from the service side, not the package running on the users computer. It's from code running on the API server! So the user can't do anything about it, and can't even inspect the code that's causing the error. This caused confusion in the past, so we changed a server config to stop forwarding backtrace information.

Issue #455, opened 2026-01-30
Use `pkgdown.yml` workflow and remove `docs` directory


peterdesmet:
With etn running locally, I think we finally switch to building the pkgdown website via GitHub actions. 🎉 

- [x] Add `pkgdown.yml` workflow
- [x] Pass `ETN_USER` and `ETN_PWD` as repository secrets (how is this currently done for R-CMD-check.yaml`?)
- [x] Build website locally to see that examples run
  - [x] `get_acoustic_detections()` has two failing examples (see comment below)
  - [x] `download_acoustic_dataset` has `dontrun`
  - [x] `write_dwc()` does not have example -> Create todo issue (cf. [movepub::write_dwc()] (https://inbo.github.io/movepub/reference/write_dwc.html))? See #463
  - [x] `list_` functions do not have examples (except for `list_values()`) -> Create issue? See #462
  - [x] All other functions have working examples
- [x] Build website locally to see that vignettes run
- [x] Remove the `docs` directory
- [ ] Switch build source to `gh-pages` (in repo settings)
- [x] Verify website
- [x] (potentially) enable [development website](https://pkgdown.r-lib.org/reference/build_site.html#development-mode) **Decided not to do this, see below**

peterdesmet:
The examples for `get_acoustic_detections()` showed two `ODBC failed` errors when running locally. @PietrH any idea what the cause is?

<details markdown="1">

```R
# Get limited sample of acoustic detections
get_acoustic_detections(limit = TRUE)
#> ℹ Preparing 
#> ✔ Preparing : will fetch 100  detections [9ms]
#> 
#> Error in get_acoustic_detections_page(): ODBC failed with error 00000 from \
#>              .
#> 
#> In call:
#> opencpu:::rapachehandler()
#> 
#> Backtrace:
#>     ▆
#>  1. └─watcher$push(cnd)
#>  2.   └─handler$error(value)
#> 
#> ℹ This is an error forwarded via the API.

# Get all acoustic detections from a specific animal project
get_acoustic_detections(animal_project_code = "2014_demer")
#> ℹ Preparing 
#> ✔ Preparing : will fetch 235.81 k detections [15s]
#> 
#> Getting detections. ■■■■■■■■■■■                       33% [6s] | ETA: 12s
#> Getting detections. ■■■■■■■■■■■■■■■■■■■■■             67% [11.6s] | ETA:  6s
#> ℹ Wrapping up
#> ✔ Wrapping up [248ms]
#> 
[...]

PR #456, opened 2026-01-30
Use pkgdown GitHub action

peterdesmet:
See #455

Hard to test this one, since GitHub action will only kick in when merging with `main`. We could try adding `v2.3.1` to:

https://github.com/inbo/etn/blob/14d035a7c60cdb5991e44cc774fb0a9931642db4/.github/workflows/pkgdown.yaml#L5

Merging should then trigger the action and build to `gh-pages`, but still serve the site from `main/docs`. Note that credentials are currently not set up.


.github/workflows/pkgdown.yaml
```diff
@@ -0,0 +1,49 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
```
PietrH:
```suggestion
    branches: [main, master, v2.3.1]
```

.github/workflows/pkgdown.yaml
```diff
@@ -0,0 +1,49 @@
...
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::pkgdown, local::.
+          needs: website
+
+      - name: Build site
+        run: pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
+        shell: Rscript {0}
```
PietrH, line 41:
```suggestion
        shell: Rscript {0}
        env:
          ETN_USER: ${{ secrets.ETN_USER }}
          ETN_PWD: ${{ secrets.ETN_PWD }}
```

Issue #457, opened 2026-01-30
Review new vignettes


peterdesmet:
- [x] Review etn_credentials
- [x] Review etn_fields
- [x] Review package-options

PR #459, opened 2026-01-30
Review authentication vignette

peterdesmet:
Related to #457

- Suggest 1 (recommended) way to register (not contact form). Should at some point be updated in README
- Use same instructions as in NEWS.md (but without the migration)
- Don't suggest `Sys.setenv()`
- Add password recovery section
- Remove author

This PR goes into the branch that already has the `authentication vignette`.

### DIFF
--- a/vignettes/authentication.Rmd
+++ b/vignettes/authentication.Rmd
@@ -1,10 +1,10 @@
 ---
-title: "Set Up Your ETN Username and Password"
+title: "Configure credentials"
 output: rmarkdown::html_vignette
 author: Pieter Huybrechts
 date: "2025-07-22"
 vignette: >
-  %\VignetteIndexEntry{Set up your ETN username and password}
+  %\VignetteIndexEntry{Configure credentials}
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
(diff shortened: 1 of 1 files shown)


PietrH:
I'm not expecting test failures here. Rerunning.

PietrH:
The registration form was broken in the past, and at that time the contact form was the only way to get an etn account. It's probably a good idea to double check the registration form works before removing the information about the contact form.

PietrH:
> I'm not expecting test failures here. Rerunning.

Probably tests need to be adapted in #461

vignettes/authentication.Rmd
```diff
@@ -16,30 +15,45 @@ knitr::opts_chunk$set(
...
+
+You may still have them stored in the [LifeWatch.be RStudio server](http://rstudio.lifewatch.be/) environment:
+
+```{r, eval=FALSE}
+Sys.getenv("userid")
+Sys.getenv("pwd")
+```
 
-Finally, you could also set them up manually using `Sys.setenv()`, but this will be reset if you restart your R session.
+Otherwise, use the [password reset form](https://rshiny.lifewatch.be/account?p=lostpass).
```
PietrH, line 59: Asked Claudia, she is discussing with Stijn.

PietrH (approved):

Issue #454, opened 2026-01-30
Compare branches to document changes in NEWS.md


peterdesmet:
@PietrH I would do a quick review of the files changed between the `main` and `opencpu` branch to see if anything **major** is missing in the NEWS.md

Todo

- [x] Verify https://github.com/inbo/etn/compare/main..opencpu
- [x] Verify https://github.com/inbo/etn/compare/v2.2.2..main
- [x] Import `purrr`, remove `R/import-standalone-purrr.R` #466 
- [x] Merge `tests/testthat/helper-expectations.R` write `tests/testthat/helper.R`
- [x] Rename `tests/testthat/test-api_helpers.R` to `test-utils-api.R`
- [x] Add `precompile.R` and run
- [x] Remove `etn_fields.Rmd`: https://github.com/inbo/etn/pull/468

NEWS.md

- [x] Mention the new vignettes.
- [x] Mention precompile.R
- [x] Mention `expect_protocol_agnostic()` in miscellaneous
- [x] Mention `fixtures`
- [x] Mention removal of `etn_fields.Rmd`

PR #466, opened 2026-02-03
Add purrr

peterdesmet:
Suggested in #454.

Rather than standalone approach which might confuse us in the future.


peterdesmet:
Note, by using backticks rather than `[]` we also resolve `devtools::document()` warning:

> utils-api.R:208: @details Could not resolve link to topic "conduct_parent_to_helpers" in the dependencies or base packages.

PietrH (approved):

PR #484, opened 2026-02-12
Skip examples or tests that need credentials if not available

PietrH:
Some examples or tests need credentials to be stored in order to run. This is no big deal but will break CRAN submission or using `devtools::check_win_release()`. This PR gets this out of the way. 

I add two helpers, one to check if credentials are set, this is used for conditional examples and for the other function: a testthat skip helper that skips a test if no credentials are set.

It should be easy to adapt these helpers to keyring later on. They may need no modification whatsoever. 

Possible followup: Some tests are already skipped if there is no local database connection, in fact, these should also be skipped if there are no credentials set. I suppose I could repeat the skip or call it from within `skip_if_not_localdb()`.

---

This is a nice to have, and not a blocker for the ETN API release.


PietrH:
There is no issue for this PR

PietrH:
The package is also too big to be sent via `devtools::check_win_release()`, i removed the fixtures, built the package locally and uploaded via the [webform](https://win-builder.r-project.org/upload.aspx); I'm not sure if it worked.

PietrH:
This seems to have a very negative effect on package coverage, does covr somehow skip these tests as well? It should have access to credentials since it worked just fine before!

PietrH:
CI also has much more skips than expected, something is wrong with my helper and covr caught it!

man/get_acoustic_detections.Rd
```diff
@@ -64,6 +64,9 @@ Get data for acoustic detections, with options to filter results. Use
 \code{limit} to limit the number of returned records.
 }
 \examples{
+\dontshow{if (etn:::credentials_are_set()) withAutoprint(\{ # examplesIf}
+# example code
```
peterdesmet: Is this the literal comment `# example code`? If yes, then I would remove.

peterdesmet: I'm also curious how

```
dontshow{if (etn:::credentials_are_set()) withAutoprint(\{ # examplesIf}
```

Renders on the pkgdown website.

tests/testthat/helper.R
```diff
@@ -109,6 +109,16 @@ skip_if_http_error <- function(url) {
   )
 }
 
+#' Skip the test if authentication credentials are not set. #
```
peterdesmet: Remove the trailing `#`

tests/testthat/helper.R
```diff
@@ -109,6 +109,16 @@ skip_if_http_error <- function(url) {
...
+#'
+#' @family helper functions
+#' @noRd
+skip_if_no_authentication <- function(){
+  testthat::skip_if_not(credentials_are_set(), message =
```
peterdesmet: Linter (and me) will probably prefer this to be written as:

```R
testthat::skip_if_not(
  credentials_are_set(),
  "No credentials are stored"
)
```

tests/testthat/test-download_acoustic_dataset.R
```diff
@@ -1,3 +1,5 @@
+skip_if_no_authentication()
```
peterdesmet: Does this need to be done outside a `test_that()` function?

peterdesmet: Same for other occurrences where the function is called at the beginning of the test file

peterdesmet (approved):
Some minor suggestions

tests/testthat/test-download_acoustic_dataset.R
PietrH: No, but this has the advantage of skipping all tests of the `test-*` file without me having to repeat it for every test. But it has the disadvantage of not counting the number of tests that are skipped. 

I took the easiest route most of the time, do you prefer me switching to skipping in the `test_that({})` context every time?

peterdesmet: Yeah, I think it's probably better to skip within the test (best practice).

man/get_acoustic_detections.Rd
PietrH: image

PietrH: Because the system that renders the pkgdown website has access to credentials (CI, or dev machine) it'll just run the examples. CRAN for example, will not.

Note the progress is still reported even though the function is not called interactively, maybe I should stop that from happening.

tests/testthat/test-download_acoustic_dataset.R
PietrH: Alright, I'll switch that around and let you know.

PietrH:
Tested on the rstudio server to ensure it also skips when it's supposed to skip when there is db access. Made a few changes in the hopes to fix #494

PR #495, opened 2026-02-26
Ignore codemeta when building pkg

peterdesmet:
I also ran `codemetar::write_codemeta()` to include more metadata, but retained the 0.3.0 (rather than 0.3.0.9000) version number.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -12,3 +12,4 @@
 ^CITATION\.cff$
 ^\.github$
 ^LICENSE\.md$
+^codemeta\.json$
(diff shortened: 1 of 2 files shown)


PietrH (approved):
Thank you for tackling this! 

Is there anything I could do to get sensible defaults for `.Rbuildignore`? I'm thinking about something like `usethis::git_vaccinate()` but for `.Rbuildignore` instead of `.gitignore`.

peterdesmet:
Although many usethis functions will also edit that file, I don't know of a blanket fix-all function.

PR #499, opened 2026-03-04
Add favicon

peterdesmet:
Adds a favicon to the pkgdown website, using https://pkgdown.r-lib.org/reference/build_favicons.html.

- I've manually edited the favicon.ico, since the output was cropped.
- I've also moved `_pkgdown.yml` to the newly created directory `pkgdown`, so everything pkgdown-related is together (same practice is used in the pkgdown repo)
- I've updated gitignore, Rbuildignore is still ok


PietrH:
Does this mean that if I were to run `pkgdown::build_favicons(pkg = ".", overwrite = TRUE)` the resulting favicons would be misalligned? Is this something we should change in the package logo?

.gitignore
```diff
@@ -37,7 +37,6 @@ rsconnect/
 
 # pkgdown website
 docs
-pkgdown
```
PietrH, line 40: What does this do?

PietrH (approved):

.gitignore
peterdesmet: It makes sure that we don't ignore the new `pkgdown` directory. We need it now, since it contains the favicons and `_pkgdown.yml`

peterdesmet:
> Does this mean that if I were to run pkgdown::build_favicons(pkg = ".", overwrite = TRUE) the resulting favicons would be misalligned? Is this something we should change in the package logo?

If by misalligned you mean "`favicon.ico` gets truncated", then yes. This is an issue with the `pkgdown::build_favicons()` function. One that also affects pkgdown itself:

1. Open https://pkgdown.r-lib.org/ in Chrome.
2. Notice how the logo in the tab has its top and bottom cut off

It is why I created the favicon.ico for etn manually, but maybe we should report it at pkgdown.

PietrH:
I can confirm this is also the case for Firefox on my system. Maybe we should report it. Maybe we could even place a PR? I suppose you did it manually with inkscape or something similar?

PR #502, opened 2026-03-11
Rename VLIZ to ETN RStudio server

peterdesmet:
And use new URL https://rstudio.europeantrackingnetwork.org. It's an alias to the same server, but good to advertise the correct name and URL.

I've also updated the release notes on:

- GitHub: https://github.com/inbo/etn/releases/tag/v3.0.0
- Zenodo: https://zenodo.org/records/18757976
- the blog post: https://oscibio.inbo.be/blog/etn-3-0-0/

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,9 +6,9 @@
 
 ## Use etn on your computer 🎉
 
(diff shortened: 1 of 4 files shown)


PietrH:
I made a small adjustment, the automatic protocol selection is dependent on the ending of the nodename. Since the new RStudio server isn't on the vliz domain, it would fail this check and fall back on the API even when a local database connection is available. 

With a small change, the package now also expects clients with nodenames ending on `europeantrackingnetwork.org` to have local database access.

PietrH (approved):

PR #518, opened 2026-03-26
Add `citation` argument to `get_x_projects()`

### LINKED ISSUE #388: Return acoustic + animal project citations
## Goal

From STRAITS Quotation:

> Add a functionality to return acoustic project citations, i.e. the citations for all acoustic (= network) projects included in a dataset. These citations will be retrieved from IMIS.

## MarineINFO = IMIS API docs
https://marineinfo.org/en/products-and-services

## Todo

- [x] add jsonlite as a suggests, and use `rlang::check_installed()` to have it as an optional dependency for `get_acoustic_citations()`
- [x] use cli to build messaging
- [x] write tests
- [ ] use vcr to cache tests
- [x] add retries? user agent to requests?
- [x] handle multiple `acoustic_project_codes`
- [x] ~~handle multiple citations: recursive function?~~ Discussed: return the first one instead
- [x] Support `animal_project_code`
- [x] ~~Split into `cite_acoustic_project()` and `cite_animal_project()`~~ `cite_imis_dataset()` works for both
- [x] Add test for `cite_imis_dataset()`
- [x] Add nice error for IMIS 404: when an IMIS dataset can not be accessed
- [ ] remove non breaking spaces from citation strings: might fix encoding issue? 
- [x] Remove `get_acoustic_citations()`
- [x] Add throttling, I might be overwhelming the marineinfo api when fetching all citations of all projects
- [x] Return data.frame with identifier (animal project code or acoustic project code), citation (+doi), doi, ownershipPersonName, email, institute 
- [x] Add optional `citation` argument to `get_animal_projects()` and `get_acoustic_projects()`, default `FALSE`
- [ ] #519
- [x] Don't warn when no citation is found.
- [x] Handle projects without IMIS id silently

## Open Questions
Input: `acoustic_project_code`
Returns: `<formatted string>` -> cli, with formatting; H2 header per returned citation. Format url as a link.

- [x] ~~ What happens if you don't provide an `acoustic_project_id`? Generate all possible citations? ~~
- [x] What should this function be called? `get_acoustic_citations()`?
- [x] How can I find multiple citations for one acoustic_project_code? `othrel` ? -> both IMIS objects refer to one another.
- [x] Should I also accept animal_project_codes?
- [x] Citations as returned by IMIS are inconsistent, sometimes a DOI is included, sometimes it isn't
- [x] ~~Encoding issues in values returned from IMIS~~
- [x] Not all acoustic_projects have a citation, should I try to make one if it is missing? -> No
- [ ] IMIS doesn't seem to accept requests from reprex, Github Actions. Only certain user agents are accepted? Need to talk ask VLIZ IT.
- [x] Sometimes HTML tags are returned in the citation. I'd like to avoid HTML to cli syntax conversion.

### Solved Questions

https://www.vliz.be/nl/imis

- How to get a (formatted) citation for a IMIS dataset -> there is an IMIS API
- How to display/return this citation via the R package -> cli, @peterdesmet will provide an example
- how to query IMIS via acoustic project id:
    - How to get from `acoustic_project_code` to `dasid` datasetid -> `etnservice::get_animal_projects()`
- Can a single `acoustic_project_code` result in 2 citations (because there are two objects in IMIS?) eg. 2014_Gudena (see below)

~The IMIS dataset id is already returned as part of `projects.sql` in etnservice, but `list_acoustic_project_codes()` only returns a single column of the table it get's as the query result:~

~https://github.com/inbo/etnservice/blob/afaaf11b7651af03e6a3a1aaeba6b2c7d1feb859/R/list_acoustic_project_codes.R#L29~

PietrH:
In this PR I've added a helper `cite_imis_dataset()` that handles parallel requests to MarineInfo/IMIS and parses the output to a rectangular format. This is actually quite a hassle because there are a lot of exceptions in the data. 

This helper is called by `get_acoustic_projects()` and `get_animal_projects()` to add a few columns with citation information when the user sets the argument `citation = TRUE`.

In the end I decided to keep all functionality together in one big helper. I could split off the parsing and the request logic. That's not too much hassle but I didn't see much short term benefit. 

Tests use cached HTTP responses by default because MarineInfo/IMIS takes a while to respond. This is also why I don't add citation information by default.

I'm very open to all feedback! 

## Todo

- [x] Update NEWS
- [ ] ~~Solve encoding issue: &nbsp is probably the cause~~ #521
- [x] Does switching to an IMIS endpoint over MarineInfo solve the encoding issue? -> Nope
- [x] Change institute over to different ownership field, see https://github.com/inbo/etn/issues/388#issuecomment-4136250201
- [x] Improve coverage
- [ ] ~~Re enable encoding tests~~
- [ ] ~~Move IMIS/Marineinfo query code to helper? Add flexiblity to make adding/changing columns easier in the future.~~ -> Out of scope

## Review comments
- [x] Bug in `get_animal_projects(animal_project_code = "ADST-Shark", citation = TRUE)`
- [x] Consider complicating ownership logic to prefer contacts over owners
- [x] Documentation of how the contact is selected is not factually correct, it's selected on OrderNr, otherwise order of response object. Not by creator or authorship status. -> Added logic, I made it so.
- [x] Add retries, throtteling
- [x] Drop support for shared first authorship: introduces extra rows downstream.
- [x] Rename contact columns
- [x] DOI format changes
- [x] Switch contact name format to FirstName then Surname (instead of Surname, FirstName).
- [x] Extend to get_cpod_projects()
- [x] Mention get_cpod_projects() in the relevant documentation.
- [x] There's a typo (ogical) and repeated content in the documentation.
- [x] Explain qs2 file
- [x] Consider removing qs2 serialized vcr cache and either skip, or remove test.

## Encoding issue

The encoding issue is probably due to an interpretation difference in non breaking spaces between ISO-8859-1 and UTF-8. This stackoverflow thread might hold the solution: https://stackoverflow.com/questions/1461907/html-encoding-issues-%C3%82-character-showing-up-instead-of-nbsp


PietrH:
The encoding issues persist! 

```
-   [1] "Aarestrup, K.; Thorstad, EB.; Koed, A.; Svendsen, JC.; Jepsen, N.; Pedersen, MI.;  Økland, F.; (2004) Acoustic receiver array in the Gudena River, Denmark, from 2005-2005. doi:10.14284/735."                     
  +   [1] "Aarestrup, K.; Thorstad, EB.; Koed, A.; Svendsen, JC.; Jepsen, N.; Pedersen, MI.;  Ã\u0098kland, F.; (2004) Acoustic receiver array in the Gudena River, Denmark, from 2005-2005. doi:10.14284/735."               
```

I believe they are due to the inclusion of non breaking spaces.

peterdesmet:
For completeness, can you also add this to argument to `get_cpod_projects()`?

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+        order_by = .data$OrderNr,
+        # Support shared first authorship
+        with_ties = TRUE
+      )
+    }) |>
+    purrr::map(\(ownership_df) {
+      dplyr::mutate(
+        ownership_df,
+        # Support missing fields, fall back to NA.
+        name = stringr::str_c(.data$Surname, .data$Firstname, sep = " "),
```
peterdesmet: I would switch this around to FirstName then Surname

tests/fixtures/citations-all.qs2
peterdesmet, line 1: What is this file for?

R/get_animal_projects.R
```diff
@@ -4,6 +4,13 @@
 #'
 #' @param animal_project_code Character (vector). One or more animal project
 #'   codes. Case-insensitive.
+#' @param citation Logical. If `TRUE`, add citation information from IMIS/MarineInfo for each
+#'  project, based on the `imis_dataset_id` field. This will add the following columns to the output:
+#'  - `citation`: A formatted citation with DOI if available.
+#'  - `doi`: The DOI for the dataset if available.
+#'  - `name`: The contact person, usually the first author. If no contact person is entered, the first author with status creator.
```
peterdesmet: Please rename these columns to:

- `contact_name`
- `contact_email`
- `contact_affiliation`

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
+#' Get the citation and associated first author information for an IMIS dataset.
+#'
+#' This function returns the the citations and some information for the first
+#' author for a dataset as stored on [IMIS](https://www.vliz.be/nl/imis) These
+#' citations will be retrieved from
+#' [MarineInfo](https://https://marineinfo.org).
+#'
+#' @inheritParams get_acoustic_projects
+#' @param imis_dataset_ids A vector of IMIS dataset ids as returned by
+#'   `get_acoustic_projects()` or `get_animal_projects()`.
```
peterdesmet: Maybe list `get_cpod_projects()` here as well.

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+#' @param warn Logical. If `TRUE`, return a warning if any of the API requests
+#'   fail, with the IMIS dataset id and error message for each failed request.
+#'   Default is `FALSE` to avoid overwhelming users with warnings when querying
+#'   many datasets, but can be set to `TRUE` for troubleshooting.ogical. If
```
peterdesmet: Typo: `ogical` and repeated content.

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+#'   be set to `TRUE` for troubleshooting.
+#' @param progress Logical. If `TRUE`, show a progress bar for the API requests.
+#'   Default is `TRUE`, but will be automatically set to `FALSE` when testing to
+#'   avoid cluttering testthat output.
+#'
+#' @returns A data.frame with 5 columns:
+#'  - The `imis_dataset_id`
+#'  - A formatted `citation` with DOI if available.
+#'  - The `doi`.
+#'  - The contact person, usually the first author. If no contact person is
```
peterdesmet: Indicate is `contact_name`

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+#'  entered, the first author with status creator.
+#'  - The corresponding `email`.
```
peterdesmet: Rename to `contact_email`

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+#'  - The corresponding `institute`.
```
peterdesmet: Rename to `contact_affiliation`

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+      ))
+    })()
+
+  # Parse the Citation and DOI ----------------------------------------------
+
+  # Avoid missing global variable R CMD NOTE
+  doi <- NULL
+  citation <- NULL
+
+  # If there is no citation, return an empty string. If there is a doi, append
```
peterdesmet, line 125: Please append the DOI to the citation as (full url, no closing period):

```
https://doi.org/<doi>
```

Not as:

```
doi: <doi>.
```

R/cite_imis_dataset.R
```diff
@@ -0,0 +1,238 @@
...
+                  !stringr::str_ends(citation, stringr::fixed(".")),
+                yes = ".",
+                no = ""
+              ),
+              doi_prefix = ifelse(doi != "", " doi:", ""),
+              doi_suffix = ifelse(doi != "", ".", ""),
+              .na = NA_character_
+              )
+          ),
+        doi = purrr::pluck(dataset_metadata, "dois", "DOI", .default = NA)
```
peterdesmet: I think it's more useful to the user to return this as full URLs, so:

```
https://doi.org/10.14284/605
```

Not:

```
10.14284/605
```

Since it's the same in the citation, you might want to add the prefix earlier.

peterdesmet (commented):
See comments for naming the comments and updating the DOI format.

Also, I noticed that for at least one project, the number of rows is  increased when asking for the citation. Maximum 1 citation/contact should be added:

``` r
library(etn)
get_animal_projects(animal_project_code = "ADST-Shark")
#> # A tibble: 1 × 11
#>   project_id project_code project_type telemetry_type project_name start_date
#>        <int> <chr>        <chr>        <chr>          <chr>        <date>    
#> 1        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> # ℹ 5 more variables: end_date <date>, latitude <dbl>, longitude <dbl>,
#> #   moratorium <lgl>, imis_dataset_id <int>
get_animal_projects(animal_project_code = "ADST-Shark", citation = TRUE)
#> # A tibble: 6 × 16
#>   project_id project_code project_type telemetry_type project_name start_date
#>        <int> <chr>        <chr>        <chr>          <chr>        <date>    
#> 1        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> 2        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> 3        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> 4        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> 5        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> 6        666 ADST-Shark   animal       ADST           ADST-Shark   2018-07-01
#> # ℹ 10 more variables: end_date <date>, latitude <dbl>, longitude <dbl>,
#> #   moratorium <lgl>, imis_dataset_id <int>, citation <chr>, doi <chr>,
#> #   name <chr>, email <chr>, institute <chr>
```

<sup>Created on 2026-04-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/cite_imis_dataset.R
PietrH: Done

peterdesmet (changes requested):

PietrH:
> For completeness, can you also add this to argument to `get_cpod_projects()`?

Done

tests/fixtures/citations-all.qs2
PietrH: This is a cached http response, but instead of using a yaml serializer, I'm using [qs2](https://github.com/qsbase/qs2), as in: https://docs.ropensci.org/vcr/news/index.html?q=qs2#new-features-2-0-0

> New serializer option qs2, using the qs2 package, generating compressed binary cassette files that are smaller than YAML or JSON files. compressed will have the greatest proportional disk space savings as cassettes have more data in them.

We can drop the whole test, it takes a while without caching (it generates all available citations). But I like it as it surfaced some bugs before.

PietrH: This is also why `{qs2}` is now in `Suggests`

PietrH:
Thank you for the review Peter. Let me know when you are done and I'll start implementing your suggestions.

tests/fixtures/citations-all.qs2
PietrH: I got rid of it.

PietrH:
I made all requested changes. You can merge if you don't feel a second review is necessary. I also messed with the logic on how to select the contact person a bit, handling more edge cases.

peterdesmet (changes requested):
I have the impression something seems to be going wrong with the citation. It's empty more often than before. E.g. it's empty for `2014_demer`.

PietrH:
> I have the impression something seems to be going wrong with the citation. It's empty more often than before. E.g. it's empty for `2014_demer`.

Added some extra tests, and fixed the bug in the string building.

PietrH:
Tests are dependent on the specific IMIS datasets not changing (adding or removing dois would break things). 

I see 2 solutions, both adding complexity to the tests:

1. Check if the dataset we are testing on is still applicable before the actual test (add expectations)
2. Search for an applicable example dataset before the test, and then test on that.

peterdesmet (approved):
Looks great, @PieterjanVerhelst should be happy with this new functionality.

@PieterjanVerhelst, you can try this out with:

```R
get_animal_projects(citation = TRUE)
```

PR #541, opened 2026-04-14
Add http skips

PietrH:
I noticed a few tests are dependent on the API being online, but will not properly skip if it is offline. I added a few extra skips. 

It's probably better on the long term to cache the HTTP responses, that way the tests are both independent on the API being online, but can still run in that case and be tested on the r-universe CI. This is already the case for many of the tests, but not all of them, causing sporadic test failures due to timeouts and such. 

This PR is the easy option, we can also just go the caching option strait away.

I've added http caching for `get_tags()` because the test file was running code outside of tests.


peterdesmet (approved):
Cursory look, looks fine.

PR #544, opened 2026-04-14
Create `get_package()`

### LINKED ISSUE #511: New function `get_package()`
Brainstormed during meeting 24/03

## Input

- [x] animal_project_code

```R
get_package(animal_project_code)
```

## Functionality

- [x] Download all acoustic data related to an **animal project** (functionality now in [`download_acoustic_dataset()`](https://github.com/inbo/etn/blob/main/R/download_acoustic_dataset.R))
- [x] Create frictionless data package
- [x] Attach one resource for each table:
  - [x] animals
  - [x] tags
  - [x] detections
  - [x] deployments
  - [x] receivers
- [x] Set `package$name` to `lower(animal_project_code)`

## Return

- [x] frictionless data package

```json
{
  "name": "2014_demer",
  "id": "https://doi.org/10.14284/432",
  "resources": [
    {
      "name": "animals",
      "path": "animals.csv"
    },
    # etc.
}
```

## Don't forget

- [x] Update the `package` parameter in `write_dwc()` to say:

    > @param package A Frictionless Data Package of ETN data, as returned by [read_package()] or [get_package()].

### LINKED ISSUE #537: Add `id = DOI` via `get_package()` and use it in `write_dwc()`
## get_package()

- [x] Use `get_animal_projects(animal_project_code = "input", citation = TRUE)` to obtain both `imis_dataset_id` and `doi`.
- [x] Set `package$id` (first parameter)
    - To `doi` if it is not `NA`
    - Otherwise, set to `https://marineinfo.org/id/dataset/<imis_dataset_id>`
- [x] Add test to verify `id` is always set

## example package

- [x] Remove `profile` (we don't need it)
- [x] Add `id` and set it to `"https://doi.org/10.14284/432"`

https://github.com/inbo/etn/blob/62148612f2dfb81d32eb514c5f0daf581d45c5f7/inst/extdata/2014_DEMER/datapackage.json#L2

## write_dwc

This function should not make any calls to the DB, which is why we're using what is provided in `package$id`.

- [x] Update `dataset_id` parameter to `dataset_id = package$id`
- [x] Update documentation to:
    > `dwc:datasetID`: `dataset_id`, defaulting to `package$id`.
- [x] Remove `dataset_id` from example
- [x] Remove `dataset_id` from snapshot test but expect output to be the same

### LINKED ISSUE #539: Add field descriptions `datapackage.json` in `get_package()`
As part of STRAITS, we plan to:

> Update the function `download_acoustic_dataset()` [...]. The `datapackage.json` file returned by the function will be updated to include field descriptions and units (where applicable).

This issue is about adding field descriptions.

## get_package()

Rather than doing this as part of `download_acoustic_dataset()`, it should be done in `get_package()`. I think the easiest is to use the movepub approach:

- When adding a resource, use custom `add_resource_etn()`
- For each field, pull the `description` (and later `unit`) from a source file (cf. `move2::movebank_get_vocabulary`)

## source information

Unlike movepub, we don't have an online vocabulary to pull the definition for each term. So we either:

1. Need to get it live from the database (I believe there is a table with definitions) with something like `get_term_definitions()`
2. Maintain the definitions (and units) as a csv in the package in `inst/extdata`. This will be faster and gives us a bit more control to maintain. It can also be replaced later.

Originally, we were thinking of maintaining all of this as part of the static https://github.com/inbo/etn/blob/main/inst/assets/datapackage.json, but that file will no longer be used: `get_package()` creates the `datapackage.json` on the fly (useful, because it accounts for optional resources like archival data). If we need to maintain something ourselves, I'd prefer a simple csv or yaml:

table | field | definition | unit
--- | --- | --- | ---
animals | animal_id | Unique identifier of ... | 
animals | animal_project_code | Code from which ... |

@sannegovaert @PietrH what do you think about the above suggestion?

@lottepohl @Stijn-VLIZ @CLAUMEMO what's the easiest way for us to get the definitions? Where are they stored?

sannegovaert:
- fix #511 
- fix #537 
- fix #539
  - field definitions, type, units and examples are added
  - when one of these is missing, it is automatically added as `null`, is this ok?


pkgdown/_pkgdown.yml
```diff
@@ -13,10 +13,9 @@ reference:
   - has_concept("access functions")
 - title: Work with animal project datasets
   contents:
-  # - get_package
+  - get_package
+  - has_concept("frictionless functions")
   - has_concept("transformation functions")
-  - read_package
```
peterdesmet, line 18: Not listing `read_package` and `write_package` verbosely here, removed them from the list since they are reexports. But I've now changed this, so they are listed as separate items via `has_concept("frictionless functions")`.

peterdesmet (approved):
I think this one is ready to merge! 🎉

PR #545, opened 2026-04-14
Upload field_definitions.tsv

peterdesmet:
Part of #539

Steps I used to create this first version:

1. Start from field_metadata_20260414.csv provided by @Stijn-VLIZ at https://github.com/inbo/etn/issues/539#issuecomment-4242210318
2. Remove tables that are currently not relevant (e.g. deviceReleaseRetrieval)
3. Pluralize table names: animal -> animals
4. Sort entire table on table, name
5. Parse out field_units_format_description to create:
    - unit: from units
    - example: from as in
    - comments: from type
6. Combine data with the returned columns from the 5 main functions
    - see where there are partial name and move that name to "alternative_name"
    - this adds the "type"
    - this adds extra rows for fields for which there are no definitions
7. Update column type with the info provided in "comments", e.g. it might be defined as "string" in what we get, but it should be an integer.

The end result is a CSV with:
- table
- name
- alternative_name
- definition
- type
- unit
- example
- comments

Where at least **all column names that we encounter** in the 5 main functions should be present (and some extra that we don't return).

I formatted as TSV, which is easier to copy/paste to and from a spreadsheet.

### DIFF
--- a/inst/extdata/field_definitions.tsv
+++ b/inst/extdata/field_definitions.tsv
@@ -1,0 +1,254 @@
+table	name	alternative_name	definition	type	unit	example	comments
+animals	acoustic_tag_id			string			
+animals	acoustic_tag_id_alternative			string			
+animals	age		Age of animal that carries tag, the units are specified in 'Age units'.	integer		2	
+animals	age_unit		Units of age.	string		durif index	
+animals	anaesthetic		Name of anaesthetic used if applicable.	string		Clove oil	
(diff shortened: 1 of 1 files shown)


peterdesmet:
Note: this file should eventually be curated so we can:

- Add type, definitions for fields that don't have one
- Maybe remove rows we don't use
- Add the projects table
- Add the reference table

sannegovaert (approved):

PR #559, opened 2026-04-21
Deprecate `download_acoustic_dataset()`

### LINKED ISSUE #556: Deprecate `download_acoustic_dataset()`
This function is deprecated in favour of:

```R
my_dataset <- get_package(animal_project_code = "2014_demer")
write_package(my_dataset, "my_directory")
```

The function internals can stay the same for now, but it should return a lifecycle deprecation warning with the above suggestion.

Note that the function is already listed as deprecated in the pkgdown function reference.

sannegovaert:
- fix #556


sannegovaert:
Ready for review after #544 is merged with main

peterdesmet (approved):
I looked at [case_match](https://github.com/tidyverse/dplyr/blob/main/R/case-match.R) to see how dplyr deprecates and used the same approach for `download_acoustic_dataset()` and `connect_to_etn()`

- lifecycle label first
- Then:

  > `function()` is deprecated. Use x instead, which [reasons]

- Then: rest of the original function documentation

Other changes I made:

- Add an example with `get_package()` and `write_package()`.
- Make sure deprecation warning `details` is styled
- Remove (almost) duplicate test

PR #564, opened 2026-04-23
Reduce detections page size

### LINKED ISSUE #558: Silent fail for `get_detections()`
This call silently failed:

```R
albert <- get_package(animal_project_code = "2013_albertkanaal")
# ── Getting data ──
# 
# 1. Getting "animals".
# 2. Getting "tags".
# 3. Getting "detections".
# ✔ Preparing : will fetch 5.48 M detections [7.1s]
```

Digging into the response:

```
> httr2::last_response()
<httr2_response>
GET https://opencpu.lifewatch.be//tmp/x0cdecd33170e5a/R/.val/feather
Status: 400 Bad Request
Content-Type: text/plain
Body: In memory (38 bytes)
> httr2::last_response() |> httr2::resp_body_string()
[1] "cannot allocate vector of size 43.8 Mb"
```

PietrH:
The OpenCPU endpoint is having memory allocation issues where it didn't before. I've reduced the default page sizes for `get_acoustic_detections()` to 250.000 (from 1.000.000) for large queries (1M+ rows). 

This will x4 the number of sessions and queries for large queries, but strongly reduce memory requirements on the OpenCPU host. Thus the client will notice slowdown, and the host will hopefully notice reduced memory usage. 

```r
get_acoustic_detections(acoustic_project_code = "albert")
```

Should now work. 

Failures shouldn't have been silent anyway, only on a hard R crash on OpenCPU. I have trouble replicating the silent crashing. The caller environment still needs to be fixed so the errors refer to their calling functions over `req_perform_opencpu()`. I'm leaving this for a separate PR. 

## To test

- [x] Does a 25M query actually work?
- [x] Is the slowdown acceptable? 

We could fine tune the page size a bit more, maybe more than 250k is fine. It's not much fun that 1M row pages used to be fine, but are problematic now.


peterdesmet:
I tried this and it worked:

```R
> start_time <- Sys.time()
> albert <- get_acoustic_detections(acoustic_project_code = "albert")
✔ Preparing : will fetch 25.95 M detections [48.7s]
✔ Wrapping up [18s]                                                              
> end_time <- Sys.time()
> end_time - start_time
Time difference of 34.07045 mins
```

Note: the first time I tried it my computer was idle for a while and then it failed, which I guess is to be expected:

```R
> albert <- get_acoustic_detections(acoustic_project_code = "albert")
✔ Preparing : will fetch 25.95 M detections…
Error in `httr2::req_perform()` at etn/R/utils-api.R:270:3:
! Failed to perform HTTP request.
Caused by error in `curl::curl_fetch_memory()`:
! Timeout was reached [opencpu.lifewatch.be]:
Operation too slow. Less than 1 bytes/sec transferred the last 600 seconds
```

peterdesmet (approved):
This works now!

1. The guessed time indicated 40 minutes, so as I user I was informed it would take a while.
2. The whole call (included preparation) finished in 34 minutes, so the guess was accurate.
3. It's a better user experience to wait for a while then it not working.
4. On my Mac at least, I have the impression you can't leave my computer on idle, because RStudio goes into sleep mode. That's a fine limitation.

PietrH:
35 minutes! That’s rough! 

We could try to optimise a bit for performance but at the cost of it failing depending on server load. 

Maybe I should add a user option for the page size so super users can play around with it?

PR #569, opened 2026-04-24
Update example_dataset

peterdesmet:
We made changes in `get_package()` (e.g. add definitions), but did not update the included example dataset.

In addition: a new convention is to have animal project names be lowercase (so it can be used for `package$name`). I've updated this in:

```
example_dataset(dataset = "2014_demer")
```

And the directory name `inst/extdata/2014_demer` in the package.


peterdesmet:
Note, see [this commit](https://github.com/inbo/etn/pull/569/changes/682dc91036be246d6dfa51b674e8cd1183e41698) to see the impact of pulling the example_dataset from the live database again with the updated `get_package()`.

R/example_dataset.R
```diff
@@ -20,8 +27,8 @@
 #' @export
 #' @examples
 #' example_dataset()
-example_dataset <- function(dataset = "2014_DEMER") {
-  available_datasets <- c("2014_DEMER")
+example_dataset <- function(dataset = "2014_demer") {
+  available_datasets <- c("2014_demer")
```
PietrH, line 31: Consider using `match.arg()` or `rlang::arg_match()`, the produced error also captures small typos.

PietrH (changes requested):
Tiny suggestion. Saves some lines and is easier to maintain.

PietrH (approved):
Actually, I've come round and now prefer your custom error. While `rlang::arg_match()` would still be easier to maintain, your error is nicer to the user. 

I'll leave it up to you if you still want to change it. But either is good for me!

PR #570, opened 2026-04-24
Reexport `frictionless::add_resource()`

### LINKED ISSUE #568: Re-export `frictionless::read_resource()`
Since we encourage working with Data Packages, it should be possible to easily extract a table from it (e.g. to compare with live data).

The easiest way to support this is to re-export `frictionless::read_resource()`:

``` r
library(etn)

animals_live <- get_animals(animal_project_code = "2014_demer")

animals_deposited <-
  example_dataset() |>
  frictionless::read_resource("animals") # Include this function in etn
```

An alternative would be to allow the reuse of `get_` functions:

```R
get_animals(my_dataset)
```

But that seems needlessly complicated and breaks the convention that `get_` functions retrieve data from elsewhere.

sannegovaert:
fixes #568


peterdesmet (approved):
👍

Note: for NEWS.md it's worth bundling related changes into one (updated in b4840f27a536d0e718bcd8bf77c2cd3a898cf5cf).

PR #575, opened 2026-04-28
Replace the `status` value from `get_package()` : `receivers.csv` to `<redacted_status_value>` snapshot test.

### LINKED ISSUE #574: Redact `status` field from `receivers.csv` snap
The status field keeps changing, and while we do want to include it into the receiver information (`get_acoustic_receivers()`, `get_package()`) we do not want to trigger test snapshots every time a receiver is changed from active to available, to lost etc. 

It's possible to redact values from snapshots. I've done it before, I can do it again. And I will!

PietrH:
The file snapshot for `receivers.csv` keeps triggering a failure because the data for the `status` field keeps changing, reflecting reality as receivers become active, inactive, are broken or repaired, or become lost or found. We don't want this change in status to trigger a test failure, so this PR adds a transformation to the test to replace these specific values in the file.

The old test looks like this:

https://github.com/inbo/etn/blob/8e066cedd2d91384ab93c3d2adf1f3c0d8545e14/tests/testthat/test-get_package.R#L32-L40

I'm simply adding a transformation function to repeatedly replace some values. I chose to hardcode them so an unexpected value for status will still trigger a test failure. 

```r
\(line) {
                         stringr::str_replace(line,
                                              stringr::fixed(",available,"),
                                              ",<redacted_status_value>,") |>
                           stringr::str_replace(stringr::fixed(",active,"),
                                                ",<redacted_status_value>,") |>
                           stringr::str_replace(stringr::fixed(",lost,"),
                                                ",<redacted_status_value>,"
                           ) |>
                           stringr::str_replace(stringr::fixed(",broken,"),
                                                ",<redacted_status_value>,"
                           )
                       }

```

I chose to keep it close to the test rather than abstract it to `testthat/helper.R` in the hope that it's clear to a reader of the test that I'm messing with the function output. That way I hope it'll be clear to someone inspecting the `_snaps/get_package/receivers.csv` file that this file is different from the output of `get_package()`.


sannegovaert (approved):
Good solution!
I just styled the code with styler so it is more readable.

PR #579, opened 2026-04-30
Hide deprecated functions

peterdesmet:
Following [lifecycle guidelines](https://lifecycle.r-lib.org/articles/communicate.html#deprecate-a-function), our 2 deprecated functions:

- Now have `@keywords internal` so they are no longer listed in the function references.
- Have update `@examples`, showing usage `# Before` and `# Now`.


NAMESPACE
```diff
@@ -1,7 +1,5 @@
 # Generated by roxygen2: do not edit by hand
 
-export(connect_to_etn)
-export(download_acoustic_dataset)
```
PietrH, line 4: By removing them from the namespace (by setting them as internal), you are breaking any scripts that still have these functions in use. There will be no deprecation message, because the functions will not be loaded by users.

Isn't this functionally the same as just removing the functions outright? 

``` r
library(etn)

download_acoustic_dataset("2014_demer")
#> Error in `download_acoustic_dataset()`:
#> ! could not find function "download_acoustic_dataset"

list_acoustic_project_codes(connection = connect_to_etn("a", "b"))
#> Error in `connect_to_etn()`:
#> ! could not find function "connect_to_etn"
```

<sup>Created on 2026-04-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

peterdesmet: My bad, I should not have replaced export with keywords. Will fix.

PietrH (approved):

Issue #222, opened 2021-10-12
Make station_name case insensitive?


peterdesmet:
This is currently disabled for `get_acoustic_deployments()` and `get_acoustic_detections()`, because it slows down `get_acoustic_detections()` considerably. That could maybe improved if `detections_limited` always returns `station_name` in uppercase?

@PieterjanVerhelst how important is it that this parameter is case insensitive?

PieterjanVerhelst:
Not that important imo. I think it is good practice to always use the exact names (taking into account lower and upper case) in coding.

peterdesmet:
Thanks @PieterjanVerhelst, then we will not implement case insensitive station name. Closing issue.

PR #589, opened 2026-05-21
read archival data: add `get_archival_data()` function

### LINKED ISSUE #409: Add `get_archival_data()` function
# STRAITS

> Add a new function `get_archival_data()` to retrieve processed tag archival data. The user can define tag (tag_serial_number) and data aggregation level (summarized or all). The function retrieves relevant data (files) and returns it to the user as a single data frame.

summarized: only return data from whole hours. Drop all other data points. I could easily add support for other steps in aggregation. Are we sure no averaging or similar takes place? 

- [x] fetch csv's and combine them
- [x] fetch them where? 
- [x] decompression
- [x] how does the user select the archival data, on what parameters? 
- [x] Is a etn database query necessary? Or can it be done purely client based? 
- [x] What postgres table contains the links to the csv files?
- [x] What is the best way to programmatically read the csv files without having to manage an authentication cookie. 
- [x] Read csv with readr or via httr2? -> httr2 for throttling and retries
- [x] Add lazy evaluation support via ~~duckdb~~? arrow 
- [x] Fix bug: col classes need to match while merging: add schema? 
```r
`purrr::list_rbind()` at etn/R/get_archival_data.R:49:3:
! Can't combine `..1$timestamp_utc` <datetime<UTC>> and `..2$timestamp_utc` <character>.
```
- [x]  Adjust throttling
- [x] BUG: running out of working memory when loading multliple large datasets into memory for parsing; Store onto disk and parse one by one? 
- [x] Add progress bar argument
- [x] Idea: Break up progress in two steps: fetching data, reading/parsing csv
- [x] Idea: Fetch files sequentially (big files saturate network connections), and use `httr2::req_progress()` to show downloaded bytes.
- [x] ~~Add aggregation and aggregation argument: via duckdb?~~ need to group by the floor of the datetime, and then aggregate. duckdb is much faster. Let's not take this as extra maintenance for now. 
- [x] Test on animal_project_code case sensitivitly
- [x] Test on invalid values for animal_project_code
- [x] Add caching via httr2 for return values: very stable
- [x] Add tests for col classes: hard to predict what cols we'll get back
- [x] Add `limit` argument for tests?
- [x] Add argument so users can store csv files: `path` or `dir`
- [x] Allow partial resume from previously downloaded files
- [x] Name the local files: uuid for now
- [x] BUG: case sensitivity in animal_project_code: https://github.com/inbo/etnservice/issues/154#issuecomment-4476022363

### Minimal Viable Product
- [x] Describe changes in NEWS
- [x] Add extension to local files
- [ ] Deploy `etnservice::get_archival_data_uuid()` to OpenCPU production 
- [x] Add test skips on data domain offline and on no credentials
- [x] Add documentation of returned object, what cols to expect
- [x] Decide: Should invalid values for arguments be checked on the client or via the service? Currently all on service, is this robust?, to client, but postponed see #588
### Postponed/Extend goals
- [ ] Should users be able to read local files without a database query? Without database credentials? 
- [ ] Don't allow tibble return on very large datasets? How to efficiently count rows of returned csv's? Decide purely on filesize?
- [ ] Stop users from trying to read all archival data available to them, this will be millions of records
- [ ] Add local caching of uuid values? very stable: cachem over memoise
- [x] return helpful error if a passed path (dir) doesn't exist
- [ ] Abstract opening arrow dataset from online hosted files: reusable for public data
- [ ] Abstract downloading a file with package defaults (declaring **user agent**, retries, path handling): can be reused for public data, and for get_acoustic_detections() feather and parquet files
- [x] rename `temp_file_paths`: not always temporary files  
- [ ] Abstract out extrating a uuid from a filepath: more readable
- [ ] tests: cache http responses?: faster, more consistent, less load on VLIZ, but possibly unacceptably increases package size 
## Questions

How to detect if a project has archival data -> Will be useful for `get_package()`, it'll sometimes need to get achival data, how to know that you need to get it.

## Requirements

- Add a new function get_archival_data() to retrieve processed tag archival data. 
- The user can define tag (`tag_serial_number`) ~~and `data aggregation level` (summarized or all). ~~ -> users can aggregate themselves for now, if there is demand we can add aggregation functions later on. You'd need to read all the data anyway. 
- The function retrieves relevant data (files) and returns it to the user as a single data frame.

PietrH:
- `get_archival_data()` sends a database request via `get_archival_data_uuid()` to get uuid identifiers for the files to download, these are then downloaded from lifewatch.com via httr2 to a local directory. Then read via arrow. 

- [x] Switch CI back to production deployment
- [x] Increase code coverage to hit target

## Other changes

- I've added a helper `is_writeable` as an analog to `assertthat::is.writeable` to check if we can write to a user supplied path and to provide a nicer error if not. 
- Fixed a small bug in `validate_login()` where a failure to provide credentials would refer to the wrong calling function.3
- regenerated all docs with a new roxygen2 version
- New helper `file_size()` a vectorized `file.size()` that also returns the paths as `names()`
- New helper `cli_yes()` a `usethis::ui_yeah()` alternative that allows for `glue` syntax.


PietrH:
I'm deploying a new version of etnservice to test, and will retest this branch.

PietrH:
Switched to production deployment: etnservice 0.7.0 is now deployed

PietrH:
Failed tests:

```r
══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-get_archival_data.R:6:3'): get_archival_data() returns a tibble by default ──
<archival_data_no_query_args/rlang_error/error/condition>
Error in `get_archival_data(limit = TRUE)`: cli-213-645
At least one of the following arguments must be provided: `tag_serial_number`, `animal_id`, or `animal_project_code`.
Backtrace:
    ▆
 1. ├─testthat::expect_s3_class(get_archival_data(limit = TRUE), "tbl") at test-get_archival_data.R:6:3
 2. │ └─testthat::quasi_label(enquo(object))
 3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
 4. └─etn::get_archival_data(limit = TRUE)
 5.   └─cli::cli_abort(...)
 6.     └─rlang::abort(...)
── Error ('test-get_archival_data.R:17:3'): get_archival_data() can return an arrow datasetquery ──
<archival_data_no_query_args/rlang_error/error/condition>
Error in `get_archival_data(return_as = "arrow", limit = TRUE)`: cli-213-663
At least one of the following arguments must be provided: `tag_serial_number`, `animal_id`, or `animal_project_code`.
Backtrace:
    ▆
 1. ├─testthat::expect_s3_class(...) at test-get_archival_data.R:17:3
 2. │ └─testthat::quasi_label(enquo(object))
 3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
 4. └─etn::get_archival_data(return_as = "arrow", limit = TRUE)
 5.   └─cli::cli_abort(...)
 6.     └─rlang::abort(...)

**Comment by PietrH** [comment]
Ready for review

File: NEWS.md [hunk 1]
```diff
@@ -13,6 +13,7 @@
   * [`dc:license`](http://purl.org/dc/terms/license) is set to `"CC-BY-4.0"` or `"CC0-1.0"` rather than a URL. The input for the `license` parameter has been updated accordingly.
   * [`dwc:identificationVerificationStatus`](http://rs.tdwg.org/dwc/terms/identificationVerificationStatus) has been added and is set to `"verified by expert"` for all records, since the taxon is assumed to be well-known before the tag was attached.
   * The animal sex, life stage and weight are added to an Extended Measurement Or Facts extension file (`emof.csv`), for better support with OBIS (#555).
+* New `get_archival_data()` will download and read processed tag archival data. The function has experimental features to return an out of memory Arrow dataset in order to allow the processing of very large amounts of data (#409).
```
sannegovaert:
```suggestion
* New `get_archival_data()` downloads and reads processed tag archival data. The function has experimental features to return an out of memory Arrow dataset in order to allow the processing of very large amounts of data (#409).
```

R/utils.R
```diff
@@ -320,6 +358,61 @@ etn_citation <- function() {
...
+    cli::cli_abort("No write permission for {.path {path}}.",
+                   call = call)
+  }
+
+  invisible(path)
+}
+
+#' Get file sizes for a vector of file paths
+#'
+#' This is a slight variant on the base function file.size. I've decided to wrap
```
sannegovaert: Suggestion to rewrite so there is no "I" and "me"

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+#' Fetch a table with UUID references to archival data files
+#'
+#' Archival data is stored in csv files hosted on lifewatch.com. To fetch these
+#' we must first query ETN for uuids that are needed to build the paths to read
+#' these csv files.
+#'
+#' @inheritParams get_animals
+#'
+#' @returns A tibble with a column `converted_archival_file_uuid` that contains
+#'   the UUID pointing to the archival data csv file. Other columns are:
```
sannegovaert:
```suggestion
#'   the UUID pointing to the archival data csv file. Other columns are
```

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+
+  uuids <-
+    uuid_tbl |>
+    dplyr::pull("converted_archival_file_uuid") |>
+    unique()
+
+  ## Stop if no data found --------------------------------------------------
+  if(length(uuids) == 0){
+    cli::cli_abort(
+      "No archival data found for the provided filters.",
```
sannegovaert: Based on https://style.tidyverse.org/errors.html
```suggestion
      "Can't find archival data the provided input.",
```

tests/testthat/test-get_archival_data.R
```diff
@@ -0,0 +1,338 @@
...
+    # This tag does not have archival data as of writing this test.
+    get_archival_data(tag_serial_number = "A19163"),
+    class = "archival_data_not_found"
+  )
+
+  # But it should return the remaining data if only some tags have no data
+
+})
+
+test_that("get_archival_data() returns error on invalid values", {
```
sannegovaert, line 109:
```suggestion
test_that("get_archival_data() forwards errors from the API", {
```

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+  return_as <- rlang::arg_match(return_as)
+
+  # At least one of the query parameters must be supplied
+  non_query_args <- c("path", "limit", "return_as", "progress")
+  query_args <- setdiff(rlang::fn_fmls_names(), non_query_args) |>
+    mget()
+  provided_query_args <- purrr::discard(query_args, is.null)
+  if(length(provided_query_args) == 0 && !limit){
+    cli::cli_abort(
+      c("{.strong Attempting to download all available archival data in one go is unwise.}",
```
sannegovaert:
```suggestion
      "i" = c("{.strong Can't download all available archival data in one go.}",
```

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+      "At least one of the following arguments must be provided: {.or {.arg {names(query_args)}}}."),
```
sannegovaert: And here "x", inserting before line 71.

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+    })
+
+  values_no_file_found <-
+    used_filters |>
+    purrr::imap(~ purrr::keep_at(.x, purrr::chuck(indexes_no_file_found, .y))) |>
+    purrr::compact()
+
+  if (length(values_no_file_found) > 0) {
+    cli::cli_warn(
+      c("No archival data was found for:",
```
sannegovaert:
```suggestion
      c("Can't find archival data for:",
```

tests/testthat/test-get_archival_data.R
```diff
@@ -0,0 +1,338 @@
...
+  skip_if_offline("www.lifewatch.be")
+  skip_if_no_authentication()
+
+  expect_identical(
+    get_archival_data(animal_project_code = "Lumpfish", limit = TRUE),
+    get_archival_data(animal_project_code = "LUMPFISH", limit = TRUE)
+  )
+})
+
+test_that("get_archival_data() returns error on invalid animal_project_code", {
```
sannegovaert, line 191: Group with the other test on forwarding errors from the API?

tests/testthat/test-get_archival_data.R
```diff
@@ -0,0 +1,338 @@
...
+    get_archival_data(
+      tag_serial_number = "A15757",
+      animal_id = c(226, # Animal not connected to the tag
+                    18113)
+    ),
+    class = "archival_data_not_found_for_filter"
+  )
+})
+
+test_that("get_archival_data() can write out to a path", {
```
sannegovaert:
```suggestion
test_that("get_archival_data() can write to a path", {
```

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+#' @returns Depending on the value of `return_as`, either an [arrow::Dataset] or
+#'   a data.frame with the archival data values. The resulting table has the
+#'   following columns:
+#'   - `tag_id`: The tag id of the tag that recorded the data.
+#'   - `timestamp_utc`: The timestamp of the measurement in UTC.
+#'   - `measurement_type`: The type of measurement (e.g. "Temp", "Pressure, etc.).
+#'   - `measurement_value`: The value of the measurement.
+#'   - `measurement_unit`: The unit of the measurement (e.g. "C", "Pa", etc.).
+#'   - `animal_project_code`: The animal project code of the animal that recorded the data.
+#'   - `animal_id`: The animal id of the animal that recorded the data.
```
sannegovaert:
```suggestion
#'   - `animal_id`: The animal id of the animal the tag was attached to.
```

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+#'
+#' @inheritParams get_animals
+#' @inheritParams get_acoustic_detections
+#' @param animal_project_code `r lifecycle::badge("experimental")` Character
+#'   (vector). One or more animal project codes. Case-insensitive. An animal
+#'   project can contain many gigabytes of archival data. Downloading this data
+#'   may take a while and  exceed the available RAM or Storage of your computer.
+#' @param path `r lifecycle::badge("experimental")` Character. A path to an
+#'   existing directory where the csv files will be stored. If NULL, the csv
+#'   files will be stored in a temporary directory that is deleted when R is
```
sannegovaert, line 16: is this for both tibble and arrow?

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+#'   restarted. If you want to keep the csv files, provide a path. Keep in mind
+#'   that the csv files can be quite large and may take up a lot of disk space.
+#'   If the requested files are already present in the provided path, they will
+#'   not be downloaded again.
+#' @param return_as `r lifecycle::badge("experimental")` Character. One of
+#'   "tibble" or "arrow". Whether to return the data as an in memory tibble or
+#'   an out of memory arrow dataset. The latter is recommended for large
+#'   datasets, but requires more disk space and may be slower to query. When
+#'   selecting "arrow" keep in mind that the downloaded data will remain stored
+#'   on your computer until R is restarted.
```
sannegovaert, line 26: See above question. Here it seems implied it is only for arrow.

R/get_archival_data.R
```diff
@@ -0,0 +1,341 @@
...
+#'
+#' @returns Depending on the value of `return_as`, either an [arrow::Dataset] or
+#'   a data.frame with the archival data values. The resulting table has the
+#'   following columns:
+#'   - `tag_id`: The tag id of the tag that recorded the data.
+#'   - `timestamp_utc`: The timestamp of the measurement in UTC.
+#'   - `measurement_type`: The type of measurement (e.g. "Temp", "Pressure, etc.).
+#'   - `measurement_value`: The value of the measurement.
+#'   - `measurement_unit`: The unit of the measurement (e.g. "C", "Pa", etc.).
+#'   - `animal_project_code`: The animal project code of the animal that recorded the data.
```
sannegovaert, line 37: More than 80 characters (also in other lines)

sannegovaert (changes requested):
Cool function!
I don't know much about this kind of functionality, so I focussed my review on
- whether things work as expected
- documentation
- style guides: https://style.tidyverse.org/documentation.html, https://style.tidyverse.org/errors.html

R/get_archival_data.R
```diff
@@ -0,0 +1,343 @@
...
+#' This function fetches processed archival data from tags. The data is stored
+#' in csv files hosted on lifewatch.com. Providing at least one of the following
+#' arguments is required: `tag_serial_number`, `animal_id`, or
+#' `animal_project_code`.
+#'
+#' @inheritParams get_animals
+#' @inheritParams get_acoustic_detections
+#' @param animal_project_code `r lifecycle::badge("experimental")` Character
+#'   (vector). One or more animal project codes. Case-insensitive. An animal
+#'   project can contain many gigabytes of archival data. Downloading this data
```
sannegovaert, line 12: There is quite some information in the parameter description. Consider to move this to a seperate section.

R/get_archival_data.R
```diff
@@ -0,0 +1,343 @@
...
+#'  to query the data without loading it all into memory at once. Keep in mind
+#'  that the downloaded data will remain stored on your computer until R is
+#'  restarted. If you want to keep the data, you can provide a path to an
+#'  existing directory where the csv files will be stored via the `path`
+#'  argument. See [arrow::Dataset] for more information.
+#'
+#' @export
+#' @family access functions
+#' @examplesIf interactive() & etn:::credentials_are_set()
+#'   get_archival_data(tag_serial_number = "A15757", limit = TRUE)
```
sannegovaert:
```suggestion
#' get_archival_data(tag_serial_number = "A15757", limit = TRUE)
```

R/get_archival_data.R
PietrH: It's certainly possible, just not something I would want to happen unintentionally. If you still wish to do this you do need to provide all the project ids you want to download. 

You could run `get_archival_data(list_animal_projects())` for example.

PietrH:
@sannegovaert, @peterdesmet Regarding caching, how do you think it _should_ work? 

- Don't store archival files at all after a call, which means subsequent calls are slower. 
- Store them for the session, you need to restart R to get rid of them. The OS may release them earlier, in which case they will be redownloaded when needed.
- Use a system cache, which can be invalidated based on time or space based rules. This could persist across sessions. This is the most complicated from a development point of view, but likely to be included in the public branch anyway.

Since I'm downloading to disk anyway, caching is basically free. But it's important to be clear about what is happening, and to give the user control over the caching mechanism so they can stop their drive from filling up. (The OS should intervene, but I've had trouble with this in the past.)

R/get_archival_data.R
PietrH: I regularly break the 80 column rule in roxygen2 documentation. Is this something we adhere to within our packages? Here especially for the one line per item structure.

R/get_archival_data.R
```diff
@@ -0,0 +1,343 @@
...
+  sensor_data <-
+    arrow::open_csv_dataset(
+      csv_file_paths,
+      schema = csv_schema,
+      # since we provide a schema, we should skip reading the header
+      skip = 1
+    ) |>
+    # Take the last 36 characters of the filename (length of a UUID)
+    dplyr::mutate(uuid = stringr::str_sub(
+      arrow:::add_filename(),
```
PietrH, line 247: This is a code smell, but as far as I can see. It's actually intended to be an exported function. Any advice?

PietrH: https://arrow.apache.org/docs/r/reference/add_filename.html

sannegovaert: It has keywords "internal". Following the license, can you copy it as a helper function in etn with reference to arrow?

sannegovaert:
> @sannegovaert, @peterdesmet Regarding caching, how do you think it _should_ work?

I think they should be stored, but I have no opinion choosing between option 2 and 3.

> The OS may release them earlier, in which case they will be redownloaded when needed.

So this is variable behavior?

> But it's important to be clear about what is happening, and to give the user control over the caching mechanism so they can stop their drive from filling up. (The OS should intervene, but I've had trouble with this in the past.)

You mean by adding a parameter eg `cache = "session"`?

PietrH:
> So this is variable behavior?
It follows the same pattern as base R, but can be overwritten by the OS. So if windows, macOS or whatever thinks it's time to clear some space, it'll do so. 

> You mean by adding a parameter eg `cache = "session"`?
I think following standard patterns is enough. I do not think offering different caching mechanisms is worth the extra code complexity in this case.

R/get_archival_data.R
PietrH: It's possible, but it'll quite a bit more than a single arrow function. I'd also have to look into the license to make sure we aren't breaking any rules. 

I really feel like they intend for this function to be public.

R/get_archival_data.R
PietrH: If a path is provided, the function will look there first and download to that dir if the files aren't present. The files from a provided path are never removed. This works regardless of the value of `return_as`. Files are always stored as exactly as they are returned from the server.

R/get_archival_data.R
PietrH: Turns out, currently the tibble path will remove the downloaded csv files after the call (using withr). But the arrow path will store them in `tempdir()` until cleared by R.

R/get_archival_data.R
PietrH: So this description is not correct, if you return as tibble, the files are removed from the tempdir after the call.

R/get_archival_data.R
PietrH: I don't have a good solution, I could bypass it by forcing hive style partitioning. But that'll create a bunch more inodes on the user system and slow things down for larger datasets. I could use `utils::exportfromnamespace` or define a global variable (it'll just work without a namespace because it's a dplyr verb). 

None of these solutions seem better to me than the internal namespace declaration.

PietrH:
Thank you for the review, I think I've covered everything. I'm not fully happy with the caching mechanism, but this chunk is likely to change pretty soon anyway to be in line with public archival data. 

Could you have another look?

PietrH:
Made some small polishing changes.

PietrH:
There may be a a load dependent failure on the fileserver of the archival files. I've sent the following message to Stijn: 

---

Ik krijg afhankelijk van de ingediende cookie HTTP 500 server errors wanneer ik archival csv files probeer op te vragen. 

Wanneer ik de vliz_webc=vliz_webc1 cookie gebruik faalt de request, maar wanneer ik de vliz_webc=vliz_webc2 gebruik lukt het wel. De server geeft me echter wel devolgende cookie terug:

Set-Cookie: vliz_webc=vliz_webc1; path=/

Mijn observaties zijn van  07:21 and 07:27 UTC., mijn IP ligt in Oslo momenteel. Maar ik heb dit ook al eerder vanaf INBO IP ranges voor gehad. 

De falende response is: 

HTTP/2 500 Internal Server Error
content-type: text/html; charset=UTF-8

Of volledig: 

-> GET /etn/archival-data/file/E6F4A925-DB8B-43B4-8A56-BADB105896E4 HTTP/2
-> Host: [www.lifewatch.be](http://www.lifewatch.be/)
-> user-agent: httr2/1.2.2 r-curl/7.1.0 libcurl/7.85.0
-> accept: */*
-> accept-encoding: deflate, gzip, br, zstd
->
<- HTTP/2 500
<- date: Tue, 22 Sep 2026 07:21:41 GMT
<- server: Apache/2.4.68 (Win64) mod_log_rotate/1.02
<- content-security-policy: upgrade-insecure-requests; script-src * 'unsafe-inline' 'unsafe-eval' blob:; object-src *; frame-ancestors 'self' [www.vliz.be](http://www.vliz.be/) [vliz.be](http://vliz.be/) [form.vliz.be](http://form.vliz.be/) [www.omes-monitoring.be](http://www.omes-monitoring.be/) [omes-monitoring.be](http://omes-monitoring.be/);
<- cross-origin-opener-policy: same-origin
<- strict-transport-security: max-age=31536000
<- x-content-type-options: nosniff
<- referrer-policy: no-referrer-when-downgrade
<- content-length: 841
<- content-type: text/html; charset=UTF-8
<- set-cookie: vliz_webc=vliz_webc1; path=/
<-
<< <!DOCTYPE html>
<< <html lang="en">
<< <head>
<<     <meta charset="UTF-8" />
<<     <meta name="robots" content="noindex,nofollow,noarchive" />
<<     <title>An Error Occurred: Internal Server Error</title>
<<     <style>body { background-color: #fff; color: #222; font: 16px/1.5 -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif; margin: 0; }
<< .container { margin: 30px; max-width: 600px; }
<< h1 { color: #dc3545; font-size: 24px; }
<< h2 { font-size: 18px; }</style>
<< </head>
<< <body>
<< <div class="container">
<<     <h1>Oops! An Error Occurred</h1>
<<     <h2>The server returned a "500 Internal Server Error".</h2>
<<
<<     <p>
<<         Something is broken. Please let us know what you were doing when this error occurred.
<<         We will fix it as soon as possible. Sorry for any inconvenience caused.
<<     </p>
<< </div>
<< </body>
<< </html>
Error in httr2::req_perform(requests[[1]], verbosity = 2) :
  HTTP 500 Internal Server Error.

Wanneer ik naar vliz_webc2 wordt verwezen, werkt alles. Dit lijkt vrij willekeurig. Na een tiental minuten werkte deze request wel:

curl -v 'https://www.lifewatch.be/etn/archival-data/file/E6F4A925-DB8B-43B4-8A56-BADB105896E4' \
  -H 'Cookie: vliz_webc=vliz_webc1'

Hoewel ik deze error als transitional zou kunnen instellen voor retries, lijkt me dit geen duurzame oplossing.

PR #598, opened 2026-06-26
Use correct roxygen tag `@returns` and Roxygen 8.0.0

### LINKED ISSUE #536: Clean up `@return`
Across all functions (including helpers and test helpers)

- [x] Make sure `@return` is not written as `@returns`
- [x] Start with capital
- [x] End with `.`
- [x] Indent second line with two spaces.
- [x] Close empty line before and after
- [x] Don't reference tables that don't exist, such as:

    > A vector of all unique `scientific_name` present in `common.animal_release`.

peterdesmet:
1. Use roxygen tag `@returns` (not `@return`)
2. Consistently list `@inheritParams` first (before `@param`)
3. Use Roxygen 8.0.0
4. Minor corrections

Point 1 & 2 have no affect on the build `.rd` files.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -77,4 +77,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.3
+Config/roxygen2/version: 8.0.0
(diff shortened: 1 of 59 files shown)


PietrH:
I'll create a small patch to fix the failing test, it's a value that seems to have been removed from the database.

PietrH (approved):
Thanks for picking this up Peter

Issue #491, opened 2026-02-24
Reduce package size: `fixtures` are quite big


PietrH:
A few of the cached API responses add a lot of size to the package, which is 70MB+ as of writing. Users don't need these files, we should try to avoid including them when people build the package or install it via devtools or pak. 

It's also not really necessary to include them on Zenodo. But this isn't as big an issue. 

As reported by Peter. 

Could we use `.RBuildignore` to have users avoid downloading these files? 

## Goals
A. Reduce fixture size
B. Don't include fixtures in built package

PR #602, opened 2026-06-29
Add new function: `get_bibliography()`

### LINKED ISSUE #587: `get_bibliography()`
While users can now get all citations via `get_package()`, @CLAUMEMO suggests to also support this for a dataframe of detections users got via `get_acoustic_detections()`.

## New function `get_bibliography()`

Functionality:

```R
my_detections <- get_acoustic_detections(scientific_name = "Mola mola")
get_bibliography(my_detections)
#> item                     type             citation
#> ETN                      data platform    Reubens J, ...
#> etn                      R package        Huybrechts P, ...
#> STRAITS_GIBRALTAR_ANIMAL animal project   ...
#> FISP                     acoustic project ...
#> Haringvliet2023-2026     acoustic project ...
```

**Input**

- [x] A dataframe with detections is expected, but any df with the columns `animal_project_code` 
and `acoustic_project_code` is ok.

**Error handling**

- [x] Dataframe does not contain the columns `animal_project_code` and `acoustic_project_code`.
- [x] Project codes cannot be found. This is already handled by `get_acoustic/animal_projects()`.

**Output**

- [x] A dataframe with 3 columns:
  - [x] `item`: project code (+ `ETN` and `etn`)
  - [x] `type`: what type of item it is: `data platform`, `R package,` animal project`, `acoustic project`
  - [x] `citation`: citation returned by `cite_imis_dataset()`
- [x] Rows:
  - [x] One row for the `ETN` data system, with a hardcoded citation
  - [x] One row for the `etn` R package, with a citation to the latest (non-dev) release
  - [x] One row for each animal project
  - [x] One row for each acoustic project

### todo

- [x] change formatting of `etn_citation()`
- [x] consider helper function `platform_citation("etn")` or data object `platform_citation` for `etn_ref`
- [x] investigate testing warning: `invalid 'cutoff' value for 'deparse', using default`
- [x] argument checking: reuse `check_value()`
- [x]  consider refactoring `check_value()`: error class + nicer return message: cutoff options, partial matching similar to `rlang:::stop_arg_match()` using base `pmatch()`.
- [ ] Speed up test via `{vcr}` or implement caching on `cite_imis_dataset()`
- [x] replace etn platform ref with: 
> Reubens, J., Aarestrup, K., Abecasis, D. et al. The European tracking network through time: united efforts to advance aquatic conservation in Europe. Anim Biotelemetry (2026). https://doi.org/10.1186/s40317-026-00475-z

## Updated function `get_package()`

- [x] Call `get_bibliography()` instead of the current code:
    https://github.com/inbo/etn/blob/e57debdf78d18495cdb8bd365ee332b44e90afa4/R/get_package.R#L156-L160

- [x] Rename the resource (and csv) from `references` to `bibliography`
- [x] Update the column names
- [x] Update `field_definitions.tsv`
- [x] Update function definition
- [x] Update `download_acoustic_dataset()` function definition (references -> bibliography)

    https://github.com/inbo/etn/blob/e57debdf78d18495cdb8bd365ee332b44e90afa4/R/download_acoustic_dataset.R#L7

PietrH:
- Added a new function `get_bibliography()` that creates a data.frame of references based on a detections-like input data.frame.
- Refactored `check_value()` to no longer use assertthat, and offer a shorter and more stylish message. When possible, offering suggestions for possible typo's:

> check_value("2001_Loire", list_acoustic_project_codes(), name = "acoustic_project_code")
Error:
! Can't find `acoustic_project_code`: "2001_Loire" in: "2004_Gudena",
  "2011_bovenschelde", "2011_Loire", …, "zeeschelde", or "ZRM_network". Did you mean
  "2011_Loire"?
  
- This change only affects `check_value()` calls on the client, I have not made any changes to `{etnservice}` which retains the same behavior as before this PR. This means that the messages generated by the API will remain very long (include all possible candidate codes) and will offer no suggestions. I can not inherit this function, because all OpenCPU messages are fetched as pure text, and the etn R package uses `{cli}`. 
- **Is this change to the error messaging worth mentioning in NEWS?**
 - Updated related documentation and `field_definitions.tsv`
 - The resulting bibliography now refers to the R package and the ETN data platform by it's paper. Both of these references are hard coded, for the R package only the version number automatically updates. I had a lot of trouble getting the output of `citation("etn")` to be consistent over platforms, and to get it into the shape that I wanted. So we ended up landing on hardcoding it in. 
 - **I might have to add a bit more to  `field_definitions.tsv`, what do you think?** 
 - I've used the `qs2` serializer for test fixtures (cached HTTP responses) to save space. See: https://qsbase.r-universe.dev/articles/qs2/vignette.html, related to #491

## Review comments

- [x] Avoid duplicate suggestions in `check_value()`
- [x] Implement message suggestion in `check_value()`
- [x] rename arg `x` in `get_bibliography()`
- [x] Improve documentation formulation in `check_value()`


R/get_bibliography.R
```diff
@@ -0,0 +1,123 @@
+#' Generate a bibliography for a data.frame of detections
+#'
+#' This function generates a bibliography for a data.frame containing
+#' detections, including citations for the ETN platform, the etn R package, and
+#' any animal and acoustic projects present in the input data.frame.
+#'
+#' @param x A data.frame containing at least the columns `animal_project_code`
```
peterdesmet: I think this parameter should be called `detections` rather than `x`.

R/utils.R
```diff
@@ -2,16 +2,28 @@
 
 #' Check input value against valid values
 #'
+#' This function checks if the input value(s) `x` are present in the valid
```
peterdesmet: I'd avoid starting function documentation with `This function` (I've already updated that in `get_bibliography()`).

My preference is:

> Checks ...

R/utils.R
```diff
@@ -20,18 +32,54 @@ check_value <- function(x, y, name = "value", lowercase = FALSE) {
...
+    any(dist_for_value <= max_dist)
+  }))) {
+    # We need to repeat the candidates so we can get the minimum distance
+    # for each missing value
+    closest_match <-
+      rep(y, length(missing_values))[purrr::map_int(distances, which.min)]
+    # If there are many candidates, truncate the list to 5 items for the error
+    # message.
+    cli::cli_abort(
+      "Can't find {.var {name}}: {.val {missing_values}} in: {.or {.str {candidates_col}}}.
```
peterdesmet: Two suggestions:

1. I think I would separate this over 2 lines:

```
c(
  "x" = "Can't find ...",
  "i" = "Did you mean ..."
)
```

2. The first line should be:

```
Can't find {.var {name}} {.val {missing_values}} in: {.or {.val {candidates_col}}}
```

- Removes first colon :
- Returns existing projects as (purple) values

R/utils.R
```diff
@@ -20,18 +32,54 @@ check_value <- function(x, y, name = "value", lowercase = FALSE) {
...
+  # If the value is not found, check if it's a typo by calculating the
+  # Levenshtein distance.
+  distances <- utils::adist(missing_values, y) |>
+    # Transpose the matrix so we can get the minimum distance per candidate
+    t() |>
+    as.data.frame() |>
+    purrr::set_names(missing_values)
+  # If any candidate string is less 3 transformations away from the reference,
+  # we can assume it's a typo and suggest it to the user.
+  candidates_col <- cli::cli_vec(y, list("vec-trunc" = 5))
```
peterdesmet, line 49: For:

```R
etn:::check_value(c("a", "b"), list_acoustic_project_codes(), name = "acoustic_project_code")
```

I get the same suggestions:

> Did you mean "Ars" and "Ars"?

Can this be avoided?

PietrH:
- I changed a test value in a failing test.
- `etn` is now requires R4.2 due to a change in `vcr`: switching to `yaml12` which requires R4.2 as a minimum. I've made a PR in vcr to document this, 
- Should we add to `field_definitions.tsv` ?

peterdesmet (changes requested):
Nice work!

I've:
- Updated the function title and documentation to make it easier to read. I now also link to the ETN citation guidelines.
- Updated the example to show how one can retrieve (already created) bibliography from a Data Package
- Used a simpler citation style for ETN, that is in line with the etn package. We don't control the style for the projects.
- Updated the example_package() to also include bibliography

Regarding your questions:

> Is this change to the error messaging worth mentioning in NEWS?

Yes, I've updated the NEWS to describe the changes in `check_value()`

> I might have to add a bit more to field_definitions.tsv, what do you think?

What do you want to add? Can this be done later?

TODO

For @PietrH I made some suggestions, especially regarding the formatting of the `check_value()` error message and the name of the parameter in `get_bibliography()`

For @lottepohl the [citation guidelines](https://europeantrackingnetwork.org/en/4-data-policy-permissions-citation-guidelines-and-data-use) could benefit from 3 changes:

- The Citation guidelines are part of a bigger document, but you can't deep link to the `Citation guidelines` header. Would be good if we have that anchor.
- The full URL is incredibly long
- The guide mentions "Network projects", while we've been calling it "Acoustic projects" for a while now.

R/utils.R
PietrH: This is a result of `"Ars"` being the closest match for both `"a"` and `"b"` which are both missing. This was not intentional, but has a certain elegance to it. 

It's not hard to avoid. And is not informative, so avoid it we will.

PietrH:
I implemented all suggestions. Thanks for the review Peter!

PR #612, opened 2026-08-20
Citations: parse json from HTTP response object instead of from disk.

PietrH:
Reading the json from disk for every marineinfo request breaks the vcr http response caching in the tests for `get_package()` since the temp json files don't exist in the testing env. 

This PR switches back to parsing these json response bodies directly from the response in memory.


PietrH:
The R4.1 failure is due to a snapshot. A different bug than this PR is trying to fix.

peterdesmet (approved):
Did not review in detail but looks like your simplifying code, so I'm all for it.

PR #620, opened 2026-09-04
Update `create_package()` snapshot

PietrH:
The snapshot test for `create_package()` is failing on main. Likely as a result of the new `{frictionless}` package version. 

Could you check if this looks all good?


PietrH:
Something is amiss, the snapshot changed again. I'll investigate

peterdesmet:
frictionless 2.0.0 is not out yet, so that should not affect things. (Don't merge this PR, because it assumes frictionless 2.0.0)

